PR #22, opened 2021-12-23
doc: Add Nix instructions to readme

siraben:
`fq` is packaged in Nixpkgs as of https://github.com/NixOS/nixpkgs/pull/151871, so these instructions will work within a few days (once the master channel advances.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ brew install wader/tap/fq
 yay -S fq # or fq-bin
 ```
 
+### Nix
+
+```sh
+nix-shell -p fq
+```
+
 ### Build from source
 
 Make sure you have go 1.17 or later installed.


jtojnar:
That requires Nix flakes enabled.

siraben:
@jtojnar do you recommend using `nix-shell -p fq` instead?

siraben:
Fixed to use stable Nix `nix-shell`

wader:
@siraben Thanks, I don't much about nix (but want to know more!) so let me know when and what to do.

siraben:
> @siraben Thanks, I don't much about nix (but want to know more!) so let me know when and what to do.

Sure thing. Like homebrew, Nixpkgs is a monorepo of package definitions, but users use channels, which are essentially points in the repo that have passed all the relevant tests. As such the instructions in this PR wouldn't work right away until the PR has made its way through the channels (see the [PR tracker](https://nixpk.gs/pr-tracker.html?pr=151871)). I'll check periodically until it's green then ping again 😄

wader:
👍  has been a pleasure to write it! hope it's useful and let me know if something seems strange or hard to understand, there are some murky areas in fq that has some accidental design atm.

siraben:
Rebased on master.

siraben:
Good to merge now, it's made its way to the unstable channel.

portothree:
Hey @siraben, sorry for jumping on this after it's already merged, but why are `nix-shell -p` being used in the section for installation? This is not actually installing the package, it's just spawing a new shell with the `jq` available there. The user would then need to run `nix-shell -p` everytime they want to use it.
It should instead be `nix-env -iA nixpkgs.fq` for non-NixOs and `nix-env -iA nixos.fq` for NixOS, this will make fq available everywhere in the system. Unless I'm missing something and there's a good reason to use the current command, happy to hear the reasoning.

siraben:
@portothree generally people in the Nix community want to avoid promoting `nix-env` (because to new users there can be confusing behavior with `nix-env -u` among other things), so I followed that convention here. Bringing it in line with the pacman and brew commands is probably fine, though.

wader:
Hey, i saw that @siraben updated to 0.0.3 in nixpkgs master 👍 but why does nix unstable still say 0.0.2? a bit confused with when things in master gets to unstable

siraben:
https://nixpk.gs/pr-tracker.html?pr=154449 still not in nixos-unstable yet

PR #35, opened 2021-12-27
doc: use nix-env instead of nix-shell to install

portothree:
As mentioned on https://github.com/wader/fq/pull/22#issuecomment-1001554152, `nix-shell -p` is not actually installing the package, this PR updates it to use `nix-env -iA` instead, this way it's behavior is the same as `brew install` and `yay -S`, making `fq` available system wide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ yay -S fq # or fq-bin
 ### Nix
 
 ```sh
-nix-shell -p fq
+nix-env -iA nixpkgs.fq # or nixos.fq on NixOS
 ```
 
 ### Build from source


siraben (approved):

jtojnar:
I do not think that we should switch to installation command just because other package managers do not support running software without installation.

Additionally, imperative installation is considered bad practice by many people in the community. They would either use `nix-shell` (or `nix shell` experimental interface) if they only need it occasionally, or install it declaratively.

portothree:
@jtojnar I don't want to sound rude, but people using Nix as a package manager to substitute something like brew is not really interested in the underlying reason to avoid **installing**. Nix here is being used the same way as brew, apt or yay, there's no need to enforce "running software without installation", because tha's the point, we want it to be installed and available system wide. Feel free to add a new section or more comments for a good reason to instead use `nix-shell`. Using `nix-shell -p` in a installation step is just bad UX.

wader:
Sorry i don't know much about nix yet. Would it be even more confusing to list both alternatives? @siraben any opinion? still approve?

siraben:
I'm sort of conflicted here and don't want to bikeshed too much. We generally don't want `nix-env` to be promoted too much nowadays (hence why Nix itself recommends `nix-shell` over `nix-env` when the user types an unknown command in the shell) because it has surprising behavior when the user does things like upgrading packages. I'm fine with `nix-env -iA nixpkgs.fq` though because the `-A` makes it specific enough. @jtojnar thoughts?

wader:
Did christmas and new years bring any ideas?

portothree:
Happy to go with whatever @jtojnar @siraben decides as I'm not as experienced in Nix as them.

wader:
👍  ok let's wait a bit more

siraben:
I recommend keeping the command as is. Nix users generally can infer what they need from the package name `fq` and documentation.

portothree:
Fair enough. Closing this PR, thanks for the discussion.

wader:
Thanks everyone for contributing. Feel free to reopen och create new issues.

PR #37, opened 2021-12-28
bson: Add decoder

wader:
Only supports basic types for now

PR #38, opened 2021-12-28
WIP: Avro: Add decoder

xentripetal:
WIP PR for discussion on design and api usage.

Currently only supports [Object Container Files](https://avro.apache.org/docs/current/spec.html#Object+Container+Files) files. Supports all primitive and complex types, but no logical types.


format/avro/codecs/codec.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/wader/fq/pkg/decode"
+)
+
+type CodecType int
+
+type Codec interface {
+	Decode(name string, d *decode.D)
+}
+
+func BuildCodec(s schema.SimplifiedSchema) (Codec, error) {
```
wader, line 15: A very subjective nitpick, as most decode methods seems short think i would return a function instead of interface and move all the decode function into BuildCodec (and use closures to pass config) and rename it to DecodeFnForType etc. But maybe logical types (don't know how they work) are more complicated?

format/avro/codecs/bytes.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/wader/fq/format/avro/schema"
+	"github.com/wader/fq/pkg/decode"
+)
+
+type BytesCodec struct{}
+
+func (l BytesCodec) Decode(name string, d *decode.D) {
+	// What if its a record with a field called name_len?
+	// using a struct is probably a better idea. But it makes it less usable
```
wader, line 12: I have grey hairs after thinking how to balance correctness and useability like this :)

format/avro/codecs/enum.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/wader/fq/format/avro/schema"
+	"github.com/wader/fq/pkg/decode"
+)
+
+type EnumCodec struct {
+	symbols []string
+}
+
+func (l EnumCodec) Decode(name string, d *decode.D) {
+	d.FieldScalarStrFn(name, func(d *decode.D) string {
```
wader, line 13: It would be nice use `scalar.Mapper` here then all enums would have both their "raw" (called actual atm) or their symbolic value accessible.

Something like this:

```go
func (l EnumCodec) MapScalar(s scalar.S) (S, error) {
	// some casting and bounds checks needed here
	// ActualS returns the decoded zigzag signed int64
	if t, ok := l.symbols[s.ActualS()]; ok {
		s.Sym = t
	}
	return s, nil
}

func (l EnumCodec) Decode(name string, d *decode.D) {
	d.FieldSFn(name, VarZigZag, l)
}
```

format/avro/codecs/long.go
```diff
@@ -0,0 +1,34 @@
...
+	var shift uint
+	for d.NotEnd() {
+		b := byte(d.U8())
+		value |= uint64(b&intMask) << shift
+		if b&intFlag == 0 {
+			return int64(value>>1) ^ -int64(value&1)
+		}
+		shift += 7
+	}
+	panic("unexpected end of data")
```
wader, line 23: In decoder i've used two different kinds of a panics, one kind is expected things like io error, malformed files etc and anything else (like a string) are treated as bugs and will crash fq. For expected things there are helper functions like `d.IOPanic`, `d.Errorf` and `d.Fatalf`.

format/avro/testdata/firstBlockCountNotGreaterThanZero.fqtest
```diff
@@ -0,0 +1,12 @@
+$ fq . firstBlockCountNotGreaterThanZero.avro
```
wader, line 1: use verbose?

wader (commented):
Thanks and nice work, looks good overall i think. I'm sorry the documentation is lacking a the moment. How confusing/frustrating was it to write a decoder? what do you think i should focus on when writing a decoder writing guide etc?

wader:
BTW you can use `make doc` to regenerate some .md files (requires some tools, graphviz, ffmpeg) and `make lint` to run linter.

format/avro/codecs/bytes.go
xentripetal: Do you think it would make sense to consume both the length and string into a single scalar and have the symbolic representation be the string?

format/avro/codecs/enum.go
xentripetal: Good idea, updated

format/avro/codecs/codec.go
xentripetal: The interfaces were originally from overcomplicating the codecs where I had other required functions so callers could have more context. 

I've cleaned this up and swapped them over to 
```go
func(string, *decode.D)
```

format/avro/testdata/twitter.fqtest
```diff
@@ -0,0 +1,61 @@
+$ fq . twitter.avro
```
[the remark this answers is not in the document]
xentripetal: Updated

format/avro/codecs/long.go
```diff
@@ -0,0 +1,34 @@
...
+
+// readLong reads a variable length zig zag long from the current position in decoder
+func VarZigZag(d *decode.D) int64 {
+	var value uint64
+	var shift uint
+	for d.NotEnd() {
+		b := byte(d.U8())
+		value |= uint64(b&intMask) << shift
+		if b&intFlag == 0 {
+			return int64(value>>1) ^ -int64(value&1)
```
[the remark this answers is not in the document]
xentripetal: Its mostly the same, except this can be variable length number of bytes. Same as what was used in the protobuf decoder, except without revealing the intermediate value.

https://avro.apache.org/docs/current/spec.html#binary_encoding

format/avro/codecs/long.go
xentripetal: Woops, fixed with d.Fatalf

xentripetal:
> BTW you can use `make doc` to regenerate some .md files (requires some tools, graphviz, ffmpeg) and `make lint` to run linter.

Cool! I've updated this PR with the regenerated docs

> How confusing/frustrating was it to write a decoder? what do you think i should focus on when writing a decoder writing guide etc?

It took a few rewrites to get it working and constantly digging through other decoders and pkg/decode/decode.go.

I feel like the most important areas to cover in early docs would be Scalars, Compounds, and Values. Covering both their relationships to each other and how they are interpreted in the final output. 

Additionally anything taking interface{} could use some more docs. But all-in-all I think the actual design is very well done and has a lot of extensibility. Awesome work!

format/avro/schema/schema.go
```diff
@@ -0,0 +1,198 @@
...
+				return s, fmt.Errorf("error parsing union type: %s", err)
+			}
+			if unionType.Type == UNION {
+				return s, errors.New("sub-unions are not supported")
+			}
+			s.UnionTypes = append(s.UnionTypes, unionType)
+		}
+	case string:
+		s.Type = v
+	case map[string]interface{}:
```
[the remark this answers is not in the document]
xentripetal: Heh I was looking at that before wasting so much time on writing this manually. But decided against it cause I didn't want to import anything that wasn't necessary into the project and wasn't sure how it handled recursion. Should have checked the mod file first 🙃 

Ill review it to see if it can handle recursion and update to use it if so.

format/avro/ocf.go
```diff
@@ -0,0 +1,124 @@
...
+		Description: "Avro object container file",
+		Groups:      []string{format.PROBE},
+		DecodeFn:    avroDecodeOCF,
+		Dependencies: []decode.Dependency{
+			{Names: []string{format.JSON}, Group: &jsonGroup},
+		},
+	})
+}
+
+func ScalarDescription(description string) scalar.Mapper {
```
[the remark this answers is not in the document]
xentripetal: Woops, deleted and moved to 
```go
scalar.Description
```

wader:
@xentripetal nice fixes and feesback! Im out traveling atm so will be a bit slow and reviewing, will answer longer in a bit

format/avro/decoders/array.go
```diff
@@ -0,0 +1,51 @@
...
+	if schema.Items == nil {
+		return nil, errors.New("array schema must have items")
+	}
+
+	valueD, err := DecodeFnForSchema(*schema.Items)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting decode fn for array item: %w", err)
+	}
+
+	//Arrays are encoded as a series of blocks. Each block consists of a long count value, followed by that many array
```
wader, line 21: I think all comments in fq uses "//<space>..." (unless it's go generate etc), would be nice to keep consistent

format/avro/decoders/record.go
```diff
@@ -0,0 +1,47 @@
...
+		fieldDecoders = append(fieldDecoders, fc)
+	}
+
+	//A record is encoded by encoding the values of its fields in the order that they are declared. In other words, a
+	//record is encoded as just the concatenation of the encodings of its fields. Field values are encoded per their
+	//schema. For example, the record schema
+	//	      { "type": "record",
+	//	        "name": "test",
+	//	        "fields" : [ {"name": "a", "type": "long"},
+	//	      				{"name": "b", "type": "string"} ]
```
wader, line 32: json indent in comment looks a bit weird (see gitbhub diff), i guess it has tab characters, maybe convert to space?

format/avro/ocf.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/wader/fq/format/registry"
+	"github.com/wader/fq/pkg/decode"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+var jsonGroup decode.Group
+
+func init() {
+	registry.MustRegister(decode.Format{
+		Name:        format.AVRO_OCF,
```
wader, line 16: I've been thinking about format decoders that themselves are encoding some generic data. Maybe they should have some help functions in jq to convert the decoded structure into something that it represent. For example avro could have something like this in a `avro.jq` file that it embeds (see https://github.com/wader/fq/pull/37)
```jq
# maybe need some handling of nested things
# and the _len is not decided yet
def avro_ocf_torepr:
  ( .blocks
  | map(
    ( .data
    | map(with_entries(select(.key|endswith("_len") | not))
    )
  )
```
That would make is possible to do:
```sh
$ fq twitter.avro avro_torepr
[
  [
    {
      "timestamp": 1366150681,
      "tweet": "Rock: Nerf paper, scissors is fine.",
      "username": "miguno"
    },
    {
      "timestamp": 1366154481,
      "tweet": "Works as intended.  Terran is IMBA.",
      "username": "BlizzardCS"
    }
  ]
]
```

Not sure about the `*_torepr` convetion yet, `*_tovalue` or `*_tojson` i think is confusing as value and json used for other things. `tovalue` used by fq and `tojson/fromjson` in jq is to/from jsong representation.

format/avro/testdata/twitter.fqtest
```diff
@@ -0,0 +1,50 @@
+$ fq verbose twitter.avro
```
wader, line 1: I've usually kept some comment where source test files come from or how to create them. To make it easier in the future to debug or add things

wader:
Had some time to think and review a bit, looking better and better. Will continue review later. Nice with the TODO list.

format/avro/testdata/quickstop-deflate.fqtest
```diff
@@ -0,0 +1,121 @@
...
+0x0000|4f 62 6a 01                                    |Obj.            |  magic: raw bits (valid) 0x0-0x3.7 (4)
+      |                                               |                |  meta[0:2]: 0x4-0x119.7 (278)
+      |                                               |                |    [0]{}: block 0x4-0x118.7 (277)
+0x0000|            04                                 |    .           |      count: 2 0x4-0x4.7 (1)
+      |                                               |                |      entries[0:2]: 0x5-0x118.7 (276)
+      |                                               |                |        [0]{}: entry 0x5-0x17.7 (19)
+0x0000|               14                              |     .          |          key_len: 10 0x5-0x5.7 (1)
+0x0000|                  61 76 72 6f 2e 63 6f 64 65 63|      avro.codec|          key: "avro.codec" 0x6-0xf.7 (10)
+0x0010|0e                                             |.               |          value_len: 7 0x10-0x10.7 (1)
+0x0010|   64 65 66 6c 61 74 65                        | deflate        |          value: "deflate" 0x11-0x17.7 (7)
```
wader, line 12: Is this deflate as in deflate compressed? if so you could probably uncompress and add as new buffer if that would be useful

format/avro/ocf.go
wader: Maybe the function should be just `torepr` and somehow figure if the format has a repr function

format/avro/decoders/bytes.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/wader/fq/pkg/decode"
+)
+
+type BytesCodec struct{}
+
+func decodeBytesFn(schema schema.SimplifiedSchema) (func(string, *decode.D), error) {
+	//bytes are encoded as a long followed by that many bytes of data.
+	return func(name string, d *decode.D) {
+		// What if its a record with a field called name_len?
+		// using a struct is probably a better idea. But it makes it less usable
```
wader, line 14: Other codec types are more than just one value so maybe it make sense to have as struct? But it's a fine line, i've usually decoded length prefix strings and null terminated string as one value but raw bytes i'm not sure how i would do it.

xentripetal:
Bringing this discussion out of the code block review and into the comments since makes it easier to follow imo:

> [I've been thinking about format decoders that themselves are encoding some generic data. Maybe they should have some help functions in jq to convert the decoded structure into something that it represent. ](https://github.com/wader/fq/pull/38#discussion_r776798557)

> [Is this deflate as in deflate compressed? if so you could probably uncompress and add as new buffer if that would be useful](https://github.com/wader/fq/pull/38#discussion_r777031206)

Yeah, the data field can be compressed in deflate, snappy, bzip2, xz, and zstandard.

I skipped adding support for that right now as I wasn't certain what the fq-ish way to resolve those would be. 

If you just hide the fact you decompressed it and only show the decompressed data, but then you're not seeing the real binary mapping. This also means you can't tie a field back to a range in the binary data (might be possible but would be incredibly complex), which IMO kills a lot of the coolness of fq.

I saw for some compression formats you decode the compression metadata/fields and add a `uncompressed` field containing the uncompressed data. Doing that here is completely feasible here but would make it so common commands someone would use on a certain avro file would differ from another depending on its codec.

I think it ties back to what you were suggesting around the _torepr convention, where you would have a representation/semantic layer that hides the compression details. The default representation would include it and you would have to query the uncompressed field to get the actual data.

However, there are already existing tools for interacting with the meaning of the data. It might be out of scope of this project to also add representation parsing to it. I think it brings questions of what formats deserve a representation layer? You could parse out pngs and try to display them as thats their representation, though that would be silly to do.


On the design idea of having a jq phrase to go to the representation
```def avro_ocf_torepr:
  ( .blocks
  | map(
    ( .data
    | map(with_entries(select(.key | endswith("_len") | not))
    )
  )
```

I don't have enough experience with jq to know if this can handle complex mappings/rules. Is it possible to make a generic phrase that can pull from the `data.uncompressed` field in all the blocks if `avro.codec` is defined, else just pull the `data` field? There's also the issue that avro doesn't require a struct/record be the top level schema. So the data field can  just be an array of ints or an array of complex records. Theres also an ambiguity in some data types where you could never know which was which based on just the binary output.

e.g. Map is just an array of key value pairs. You would want the representation of map to be a map/struct. But if someone manually specified a schema as an array of key value pairs, you would want the representation to be an array of key value pairs.

I believe the only way to solve this is to have a context of the schema when parsing it, trying to making a transformation ruleset defined entirely on the output would be impossible. Though I might just be overthinking it.

Thoughts?

wader:
> @wader Bringing this discussion out of the code block review and into the comments to make it easier to follow:

Good idea, sorry about that. Would be nice if github allowed "normal" comments when doing a draft review somehow.

> Yeah, the data field can be compressed in deflate, snappy, bzip2, xz, and zstandard.
> 
> I skipped adding support for that right now as I wasn't certain what the fq-ish way to resolve those would be.
> 
> If you just hide the fact you decompressed it and only show the decompressed data, but then you're not seeing the real binary mapping. This also means you can't tie a field back to a range in the binary data (might be possible but would be incredibly complex), which IMO kills a lot of the coolness of fq.

Yes the best fq can currently do for this is that it allows to have new buffers inside a decode tree, it won't allow you to map locations back in any way but it can give some hint of where the in "parent" buffer it originated from. You will notice in a decode value "dump" tree that the address column is indented per "buffer" level to give some hint.

> I saw for some compression formats you decode the compression metadata/fields and add a `uncompressed` field containing the uncompressed data. Doing that here is completely feasible here but would make it so common commands someone would use on a certain avro file would differ from another depending on its codec.
> 
> I think it ties back to what you were suggesting around the _torepr convention, where you would have a representation/semantic layer that hides the compression details. The default representation would include it and you would have to query the uncompressed field to get the actual data.
> 
> However, there are already existing tools for interacting with the meaning of the data. It might be out of scope of this project to also add representation parsing to it. I think it brings questions of what formats deserve a representation layer? You could parse out pngs and try to display them as thats their representation, though that would be silly to do.

Yeah i think your right, a representation layer only makes sense if there is an obvious way to do it and that it does not involve too much work. Maybe in some cases it's better to have per format documentation with some hint and snippets for how to iterate/traverse the structure?

I've actually joked that fq should try show you everything a normal user won't not want to see :)

> On the design idea of having a jq phrase to go to the representation, e.g.
> 
> ```
>   ( .blocks
>   | map(
>     ( .data
>     | map(with_entries(select(.key | endswith("_len") | not))
>     )
>   )
> ```
> 
> I don't have enough experience with jq to know if it can handle complex mappings/rules. Is it possible to make a generic phrase that can pull from the `data.uncompressed` field in all the blocks if `avro.codec` is defined, else just pull the `data` field?

Ignoring that we should probably not do it: yes i think so, jq the langauge is very capable so probably possible. I should really write some introduction to jq with the aim at how to use it with fq. Also i'm using https://github.com/wader/vscode-jq which makes it a lot nicer to write lot if jq code... should really clean up and package that.

> There's also the issue that avro doesn't require a struct/record be the top level schema. So the data field can just be an array of ints or an array of complex records. Finally, there's an ambiguity in some data types where you could never know which was which based on just the output. e.g. Map is just an array of key value pairs. You would want the representation of map to be a map/struct. But if someone manually specified a schema as an array of key value pairs, you would want the representation to be an array of key value pairs.
> 
> I believe the only way to solve this is to have a context of the schema when parsing it, trying to making a transformation ruleset defined entirely on the output would be impossible. Though I might just be overthinking it.
> 
> Thoughts?

Sounds like we should skip trying to make this too smart, let's first make a really good basic decoder and then see how it's used or what feels missing? Really appreciate the work and thinking you put into this!

Hope you had a great new years!

wader:
Hey, hope you'r doing great. Let me know if there was something i missed to answer or so

xentripetal:
Thanks mate, all good. Just haven't had spare time to finish this PR yet.

wader:
👍  Good to hear. No stress and i don't think my messing around in master should cause much more conflicts than it already has if you let it rest for a while

wader:
Hey, i did some cleanup/sorting related to list of formats, it probably caused some collision but after resolve them i think they will be less likely to happen

format/avro/decoders/enum.go
```diff
@@ -0,0 +1,33 @@
...
+	//	      {"type": "enum", "name": "Foo", "symbols": ["A", "B", "C", "D"] }
+	// This would be encoded by an int between zero and three, with zero indicating "A", and 3 indicating "D".
+	return func(name string, d *decode.D) interface{} {
+		var val string
+		d.FieldSFn(name, VarZigZag, scalar.Fn(func(s scalar.S) (scalar.S, error) {
+			v := int(s.ActualS())
+			if v < 0 || v >= len(schema.Symbols) {
+				return s, errors.New("enum value of out range")
+			}
+			val = schema.Symbols[v]
```
wader, line 27: Could the `Symbols` member be a new type that implements `scalar.Mapper`? would more or less be that same code but moved to a method.

format/avro/decoders/fixed.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func decodeFixedFn(schema schema.SimplifiedSchema, sms ...scalar.Mapper) (DecodeFn, error) {
+	if schema.Size < 0 {
+		return nil, errors.New("fixed size must be greater than or equal to zero")
+	}
+	size := int64(schema.Size)
+	// Fixed instances are encoded using the number of bytes declared in the schema.
+	return func(name string, d *decode.D) interface{} {
+		bb := d.FieldRawLen(name, size*8, sms...)
```
wader, line 18: Some reason this is raw and not `d.FieldU(nane, size*8, sms)` etc?

format/avro/decoders/logicaltypes.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil
+	}
+}
+
+type TimestampMapper struct {
+	Precision Precision
+}
+
+func (t TimestampMapper) MapScalar(s scalar.S) (scalar.S, error) {
+	v, ok := s.Actual.(int64)
```
wader, line 49: If you know it must be a `int64` unless there is a bug you can use`s.ActualS()` which will panic

format/avro/decoders/logicaltypes.go
```diff
@@ -0,0 +1,105 @@
...
+	case "time-nanos":
+		return TimeMapper{Precision: NANOSECOND}
+	case "date":
+		return DateMapper{}
+	default:
+		return nil
+	}
+}
+
+type TimestampMapper struct {
```
wader, line 44: Like this!

format/avro/decoders/long.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/wader/fq/format/avro/schema"
+	"github.com/wader/fq/pkg/decode"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+const intMask = byte(127)
+const intFlag = byte(128)
+
+// VarZigZag reads a variable length zigzag long from the current position in decoder
+func VarZigZag(d *decode.D) int64 {
```
wader, line 13: Error if more than 8 bytes? btw i added support for bigint in master yesterday if that is interesting for avro

format/avro/decoders/null.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/wader/fq/format/avro/schema"
+	"github.com/wader/fq/pkg/decode"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+func decodeNullFn(schema schema.SimplifiedSchema, sms ...scalar.Mapper) (DecodeFn, error) {
+	// null is written as zero bytes.
+	return func(name string, d *decode.D) interface{} {
+		// Is this the best way to represent null in fq?
```
wader, line 12: Good question, i should give this some thought, have the same issue in other decoders

wader:
It seems like `doc/file.mp4` got removed?

format/avro/decoders/null.go
wader: Now there more proper nil support #81, it was kind of support in interp but i forgot about for the decoder as non of the earlier decoded needed it.

format/avro/decoders/fixed.go
xentripetal: Fixed is a byte array of a known size. Can be larger than 8 bytes so I don't think FieldU is the correct type here. Let me know if I'm misunderstanding something here.

format/avro/decoders/null.go
xentripetal: Thanks, updated to use FieldValueNil

xentripetal:
Hmm it seems like when I run `make doc` on my end it modifies doc/file.mp4. Maybe different defaults on ffmpeg between versions?

I'm using 
> ffmpeg version 4.4.1 Copyright (c) 2000-2021 the FFmpeg developers
  built with Apple clang version 13.0.0 (clang-1300.0.29.3)

on AArch64

format/avro/decoders/enum.go
xentripetal: implemented

format/avro/decoders/logicaltypes.go
xentripetal: Nice! swapped

format/avro/decoders/long.go
xentripetal: Good idea, set it to fatal on over 8 bytes

xentripetal:
@wader Wanting your opinion on `ocf.go::decodeHeader` when you have a sec:

The header is encoded using the avro specification, so it can be decoded using the avro decoders I've added. Previously I was manually parsing the header and duplicating code for the map parsing. This gave the ability to use the JSON decoder for the schema string on the header, which will always be JSON. 

```golang
d.FieldStructArrayLoop("entries", "entry", func() bool { return i < blockCount }, func(d *decode.D) {
...
    if key == "avro.schema" {
	v, _ := d.FieldFormatLen("value", valL*8, jsonGroup, nil)
	s, err := schema.From(v.V.(*scalar.S).Actual)
	headerData.Schema = &s
    }
...
```

I've just changed it over to instead use a constant header schema and parse it using the avro decoders in the [latest commit](https://github.com/wader/fq/pull/38/commits/ee184075b7576d3e9a4a5bab70b9216ce93601fa). However this loses the ability to treat the `avro.schema` field as json and allow fq to query it without piping the output into another instance.

I think this method is a lot cleaner, but I'm not sure if its a good idea to lose this functionality for some code cleanliness. Would love to know if you have an opinion or can think of a better way to design my avro decoders to allow treating that map entry as JSON.

wader:
Nice, lots of progress. You really didn't start with the simples decoder :) impressive

About `doc/file.mp4` yes `make doc` automatically recreates it if the makefile has changed i think, maybe should change that.  And yes probably just different ffmpeg version and the file was probably not deleted just changed, sorry.

Yeah think i like the new approach for header decoding more and the fq JSON decoder is a bit of a hack at the moment, maybe we can re-add something later and keep it as a string for now? you can always use `fromjsom` in jq to parse it, ex: `fq. '.header.meta[].data[] | select(.key.data == "avro.schema").value.data | fromjson | ...' file.avro`. 

Another alternative might be to somehow add a parsed schema version as `.schema` root value? but not sure if that would be via the fq JSON decoder or add some kind of `d.FieldValue("schema", v)` thingy? 

BTW the code snippet was to show how it was done before?

About bigint and fixed type: Now i see, fixed is raw unstructured bytes and it's up to the reader to use "name" etc to know what it is? Was thinking if there were any > 64bit integers in avro then you can use bigint now.

wader:
@xentripetal Hey, let me know if you want assistance with anything, but not stress. Hope the last big changes in master didn't cause too much issues.

format/avro/testdata/quickstop-deflate.fqtest
xentripetal: Yeah this is probably the best option. How would you take the decoded byte array and use it with a decode.D?

wader: If the uncompressed data is just opaque and will not be decoded further you can do something like this i think:

```go
br := d.FieldRawLen("compressed", ...)
bb := &bytes.Buffer{}
d.MustCopy(bb, flate.NewReader(bitio.NewIOReader(br)))
d.FieldRootBitBuf("uncompressed", bitio.NewBitReader(bb.Bytes(), -1))
```

There are some helpers for doing similar things but they all seem to decode the uncompressed data somehow, like `d.TryFieldReaderRangeFormat`, `d.FieldFormatBitBuf` and `d.FieldRoot*Array/Struct/Fn`. Maybe there should be some helper for this case also? ...also really need to clean those functions up, but is a bit tricky to generelize i've noticed :)

xentripetal:
Hey sorry for the delay, got distracted with some other projects. Finished out all the WIP goals I set for myself. 

Ready for full review whenever you have the time!

format/avro/testdata/quickstop-deflate.fqtest
xentripetal: ty, used the d.FieldArrayRootBitBufFn

Doctor-love:
Would be exciting to see this and #51 land in the next release - nice work, @xentripetal !

wader:
Oh sorry i noticed now that i was commenting on a old commit

wader:
> Hey sorry for the delay, got distracted with some other projects. Finished out all the WIP goals I set for myself.
> 
> Ready for full review whenever you have the time!

Great stuff!

BTW if you want to add documentation specific to `avro_ocf` you can add a `avro_ocf.md` file and then `make doc` will include it in formats documentation. For example limitations, things not implemented etc or just examples how to use it.

format/avro/avro_ocf.md
```diff
@@ -0,0 +1,7 @@
+Supports reading Avro Object Container Format (OCF) files based on the [1.11.0 specification](https://avro.apache.org/docs/current/spec.html#Object+Container+Files).
+
+Capable of handling null, deflate, and snappy codecs for data compression.
+
+Limitations:
+ - Schema does not support self-referential types, only built-in types.
+ - Decimal logical types are not supported for decoding, will just be treated as their primitive type
```
wader, line 7: 👍  this puts the other format documentation to shame. Good idea to have links specification etc, if your using fq for something if quite likely you want to have has much details and help as possible :)

I've usually collect links and useful tools etc as comments in the go-files but much of that can probably move out to documentation.

wader:
> It does but I'm fairly sure it requires that the compressed data be in a different framing than what avro puts them as https://github.com/google/snappy/blob/main/framing_format.txt
> 
> I don't know this for certain but every time I tried to use the snappy.NewReader(...) it resulted in it erroring out with invalid data.
> 
> ## Edit
> Yeah the docs for Reader say
> 
> ```
> // Reader handles the Snappy stream format, not the Snappy block format.
> ```
> 
> and docs for Decode say
> 
> ```
> // Decode handles the Snappy block format, not the Snappy stream format.
> ```

Aha yeah that could be it, i've run into a similar issue with bzip2 (there i let the reader read a bit earlier i think). We can leave it as it is i think maybe revisit it later. Possibly add a comment about it

wader:
> Ah yeah that makes sense. Swapped to bitio.ReadFull
> 
> Missed all those functions since I normally just go off methods for feature discovery.

Yeah that is probably the best way atm so no worries. I'm sorry i haven't had time to write proper decoder documentation yet.

wader:
> Nice! wasn't aware of that feature. Yeah the full dump is over 8k lines. Swapped to truncate

Sorry again :) should also add more documentation about all options. Things have been a bit in flux so have skipped/forgotten to documenting until i felt it was stable or good enough.

wader:
Seems the fqtest fails needs an update after the UTC change

format/avro/decoders/bytes.go
```diff
@@ -0,0 +1,27 @@
...
+	// Bytes are encoded as a long followed by that many bytes of data.
+	return func(name string, d *decode.D) interface{} {
+		var val []byte
+
+		d.FieldStruct(name, func(d *decode.D) {
+			length := d.FieldSFn("length", VarZigZag)
+			r := d.FieldRawLen("data", length*8, sms...)
+
+			val = make([]byte, length)
+			if _, err := r.ReadBits(val, length*8); err != nil {
```
wader, line 20: Use bitio.Copy etc? sorry didn't notice this one before. BTW i've started to use `br` as a short name for bitio.Reader, think to not confuse with io.Reader that is usually `r`

format/avro/decoders/logicaltypes.go
```diff
@@ -0,0 +1,104 @@
...
+	s.Sym = ts.UTC().Format(time.RFC3339Nano)
+	return s, nil
+}
+
+type TimeMapper struct {
+	Precision Precision
+}
+
+func (t TimeMapper) MapScalar(s scalar.S) (scalar.S, error) {
+	v, ok := s.Actual.(int64)
```
wader, line 72: Use s.ActualS()?

format/avro/decoders/logicaltypes.go
```diff
@@ -0,0 +1,104 @@
...
+		return s, errors.New("unknown precision")
+	}
+	return s, nil
+}
+
+type DateMapper struct {
+}
+
+func (d DateMapper) MapScalar(s scalar.S) (scalar.S, error) {
+	v, ok := s.Actual.(int64)
```
wader, line 95: Use s.ActualS()?

format/avro/decoders/long.go
```diff
@@ -0,0 +1,37 @@
+package decoders
+
+import (
+	"github.com/wader/fq/pkg/decode"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+const intMask = byte(127)
+const intFlag = byte(128)
```
wader, line 9: Could use `0b111_1111` and `0xb1000_0000` to make it even more explicit? it the byte cast needed btw?

format/avro/decoders/logicaltypes.go
```diff
@@ -0,0 +1,104 @@
...
+	if !ok {
+		return s, errors.New("not an int64")
+	}
+	s.Sym = time.Unix(0, 0).AddDate(0, 0, int(v)).UTC().Format("2006-01-02")
+	return s, nil
+}
+
+// Todo Decimal: https://github.com/linkedin/goavro/blob/master/logical_type.go
```
wader, line 103: There is BigInt support now so could be used for this i think? but we can do that later maybe

xentripetal: I think it would require Big.Rat support

xentripetal:
I think its good to merge, in the future I think adding decimals and user defined types in schema decoding would be good, but not really required.

wader:
🥳  Thanks a lot for this and hope for more contributions in the future!

wader:
BTW hopefully will do a release quite soon, trying to keep it around every 2-3 weeks for now.

Issue #43, opened 2021-12-30
Mach-O Binary Support


Akaame:
ELF is already supported. I think it is reasonable to add Mach-O format as a supported binary format to this tool.

As a reference for implementation I suggest following [this document](https://github.com/aidansteele/osx-abi-macho-file-format-reference).

wader:
Hi, yes that is the plan. Im currently working on an improved elf decoder and i hope other executable formats wont be that much different. Is it something you would like to help work on?

Akaame:
I actually wanted to copy elf decoder and change necessary bits until it works for mach-o binaries (and archives) over the New Year's break. Should I wait for the improved decoder or get to work?

wader:
Ok nice! Dont thank i will work much on the elf decoder the next few days, away skiiing! You can find the elf improvments WIP PR here https://github.com/wader/fq/pull/40

wader:
The tricky part for elf is that you have to do some kind of two-pass as the string tables might come after they are used

zhangkaizhao:
I ran the `fq` cli on the `mdbook` cli and got strange output on macOS Monterey 12.1 x86_64:
```sh
uname -mrs
Darwin 21.2.0 x86_64
$ fq --version
0.0.3
$ mdbook --version
mdbook v0.4.15
$ file fq
fq: Mach-O 64-bit executable x86_64
$ file mdbook
mdbook: Mach-O 64-bit executable x86_64
$ fq . mdbook
        |00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d|0123456789abcd|.{}: mdbook (mp3)
        |                                         |              |  headers[0:0]:
0x000000|cf fa ed fe 07 00 00 01 03 00 00 00 02 00|..............|  unknown0: raw bits
*       |until 0x64ac.7 (25773)                   |              |
0x0064a0|                                       ff|             .|  frames[0:2]:
0x0064ae|e2 a8 01 0f 85 31 09 00 00 44 8b bd c8 fa|.....1...D....|
*       |until 0x8926.7 (9338)                    |              |
0x006b76|                  00 00 00 48 8d b5 30 fc|      ...H..0.|  unknown1: raw bits
0x006b84|ff ff 48 89 b5 48 fd ff ff 48 c7 85 50 fd|..H..H...H..P.|
*       |until 0x86d7.7 (7004)                    |              |
        |                                         |              |  footers[0:0]:
0x00891a|                                       a5|             .|  unknown2: raw bits
[...]

PR #51, opened 2022-01-04
[WIP] 43: Initial impl for macho support.

Akaame:
[WIP]
Several issues remain for symbol mapping for flags and constant fields
UNIX/THREAD LC .state field has to be visualized for PPC/64, ARM/64 X86-64

I have created a working barebones implementation there are several improvements to be made here:
We need to have a SymStr mapping for cpusubtype but it depends on cputype and an arbitrary number.
Flag fields for individual messages are in uint32_t, I think it is user friendly to have those expanded to see which flags are set or not.
LC_THREAD/LC_UNIXTHREAD commands have a map of Register Name to Register Value. The visualization for this must be implemented for major architectures as I have written in the commit message

It correctly parses the `fq` binary on MacOS.

Solves #43


Doctor-love:
Nice work - excited to see how this progresses

wader:
Hey, very nice start! much appreciated you working on this

> [WIP] Several issues remain for symbol mapping for flags and constant fields UNIX/THREAD LC .state field has to be visualized for PPC/64, ARM/64 X86-64

Haven't look at macho closely myself yet, will you need to do some kind of two-pass you similar to string tables for ELF? (tables and references to it can come in any order).
 
> I have created a working barebones implementation there are several improvements to be made here: We need to have a SymStr mapping for cpusubtype but it depends on cputype and an arbitrary number.

For icmp i did this https://github.com/wader/fq/blob/master/format/inet/icmp.go#L52 is that what you mean? otherwise i think you could just use plain switch/if etc and use different scalar mappers.

> Flag fields for individual messages are in uint32_t, I think it is user friendly to have those expanded to see which flags are set or not. 

If they are useful booleans i think it makes sense to expose then as separate fields in a struct.

Some general comments:
- I try to use the first line commit message convention `contexta,contextb,...: message` which i like and i think makes it easier to both write and search messages. So this commit could be something like `macho: Add basic decoder`
- I've start to use more lowercase/jq-ish strings for syms as they will probably be use conditions etc that way it makes it easier to type. What do you think?
- Tests and test files and if possible script/instructions how they were created would be nice.

Let me know if you have any other questions or API suggestions.

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+		Name:        format.MACHO,
+		Description: "Mach-O is the native executable format of binaries in OS X",
+		Groups:      []string{format.PROBE},
+		DecodeFn:    machoDecode,
+	})
+}
+
+const (
+	MH_MAGIC    = 0xfeedface
+	MH_CIGAM    = 0xcefaedfe
```
wader, line 26: You can add  `//nolint:revive` above the const if you want to shut the linter up about this. For ELF i've also kept the original const names as i think it makes sense

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+	MH_CIGAM:    32,
+	MH_MAGIC_64: 64,
+	MH_CIGAM_64: 64,
+}
+
+var endianNames = scalar.UToSymStr{
+	MH_MAGIC:    "little-endian",
+	MH_CIGAM:    "big-endian",
+	MH_MAGIC_64: "little-endian",
+	MH_CIGAM_64: "big-endian",
```
wader, line 49: I've used underscore for other formats, feels more jq-ish

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+		case LC_TWOLEVEL_HINTS:
+			d.FieldU32("offset")
+			d.FieldU32("nhints")
+		case LC_LOAD_DYLIB, LC_ID_DYLIB, LC_LOAD_UPWARD_DYLIB, LC_LOAD_WEAK_DYLIB, LC_LAZY_LOAD_DYLIB, LC_REEXPORT_DYLIB:
+			d.FieldStruct("dylib_command", func(d *decode.D) {
+				offset := d.FieldU32("offset")
+				d.FieldU32("timestamp") // TODO human readable
+				d.FieldU32("current_version")
+				d.FieldU32("compatibility_version")
+				d.FieldStrFn("name", func(d *decode.D) string {
```
wader, line 322: Could this be `d.FieldUTF8("name", int(cmdsize) - int(offset))`?

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+			d.FieldU32("nextrel")
+			d.FieldU32("locreloff")
+			d.FieldU32("nlocrel")
+		case LC_BUILD_VERSION:
+			d.FieldU32("platform")
+			d.FieldU32("minos")
+			d.FieldU32("sdk")
+			ntools := d.FieldU32("ntools")
+			var ntoolsIdx uint64
+			d.FieldStructArrayLoop("tools", "tools", func() bool {
```
wader, line 428: Second argument is name of struct in array (only shown with verbose display atm). I've usually used plural for the array and singular for each entry. So "tools" and "tool" here for example. But will be a lot of "tool" fiends here, could the "tool"  fiend in the struct be name something else, "name" etc?

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+			d.FieldStruct("uuid_command", func(d *decode.D) {
+				d.FieldRawLen("uuid", 16*8)
+			})
+		case LC_SEGMENT, LC_SEGMENT_64:
+			// nsect := (cmdsize - uint64(archBits)) / uint64(archBits)
+			var nsects uint64
+			if archBits == 32 {
+				d.FieldStruct("segment_command", func(d *decode.D) {
+					d.FieldRawLen("segname", 16*8)
+					d.FieldU32("vmaddr")
```
wader, line 259: You can do `d.FieldU32("vmaddr", scalar.Hex)` to give some number base display hints

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+		case LC_UUID:
+			d.FieldStruct("uuid_command", func(d *decode.D) {
+				d.FieldRawLen("uuid", 16*8)
+			})
+		case LC_SEGMENT, LC_SEGMENT_64:
+			// nsect := (cmdsize - uint64(archBits)) / uint64(archBits)
+			var nsects uint64
+			if archBits == 32 {
+				d.FieldStruct("segment_command", func(d *decode.D) {
+					d.FieldRawLen("segname", 16*8)
```
wader, line 258: Not a string?

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+					d.FieldU32("fileoff")
+					d.FieldU32("tfilesize")
+					d.FieldS32("initprot")
+					d.FieldS32("maxprot")
+					nsects = d.FieldU32("nsects")
+					d.FieldU32("flags") // TODO expand flags
+				})
+			} else {
+				d.FieldStruct("segment_command_64", func(d *decode.D) {
+					d.FieldStrFn("segname", func(d *decode.D) string {
```
wader, line 270: If you want to trim null:s you can use `scalar.Trim("\x00")` and if it's very common i guess you could make a helper function, but it's quite nice to keep things stright forward also

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+		} else {
+			archBits = 64
+		}
+	} else {
+		d.Fatalf("Invalid magic field")
+	}
+
+	d.SeekAbs(0)
+	d.FieldStruct(fmt.Sprintf("mach_header_%d", archBits), func(d *decode.D) {
+		d.FieldU32("magic", magicToHex, classBits, endianNames)
```
wader, line 230: I think you can replace `magicToHex` with `scalar.Hex`. 

Maybe can use `UToScalar` and use description? other alternatives is to use a struct or use `d.FieldValue*` if you want to add for a lack of better name "synthetic" values that don't have a bit range.

Akaame:
Hey, thanks for taking such a thorough look at it. I admit that I did not take time to polish the code and it shows (wanted to get a working version out of the gate). Will try to work on your suggestions and comments before the EOW.

Cheers.

wader:
@Akaame 👍  no problem and good to idea to create an early PR i think. I will probably add more random comments as i come up with things. But feel no stress working on it and feel free contact me if you have questions or feel stuck at something.

Akaame:
>Haven't look at macho closely myself yet, will you need to do some kind of two-pass you similar to string tables for ELF? (tables and references to it can come in any order).

No I need to make a huge table of all register values for that architecture and use it to decode `state` field, I am pushing that off for now as it is too much menial work 😄 

> For icmp i did this https://github.com/wader/fq/blob/master/format/inet/icmp.go#L52 is that what you mean? otherwise i think you could just use plain switch/if etc and use different scalar mappers.

🤦 why did I not think of this?? Thank you.

> I try to use the first line commit message convention contexta,contextb,...: message which i like and i think makes it easier to both write and search messages. So this commit could be something like macho: Add basic decoder

Amended and followed that structure for following commits.

> I've start to use more lowercase/jq-ish strings for syms as they will probably be use conditions etc that way it makes it easier to type. What do you think?

This is the general structure I followed if I got what you meant right.

e.g: `./fq '.load_command[9].dylib_command.compatibility_version' fq`

> Tests and test files and if possible script/instructions how they were created would be nice.
You can use make lint to run linter locally, make doc to generate documentation and other things. When you add tests you can generate the "actual" output, see dev documentation.

For test I will look at the tests for another format and adopt them.
make lint fails for `cpuSubTypes` It will be fixed.
Ran make doc.

format/macho/macho.go
Akaame: The first paragraph is addressed.

I did not get what you meant by the second one.

format/macho/macho.go
Akaame: If this is about segname I am using FieldUTF8NullFixedLen for this line too.

format/macho/macho.go
Akaame: This is embarrassing but at the start I could not get the hang of it. Sad to see this being trickled down. Fixed

format/macho/macho.go
Akaame: This and similar `addr` fields are updated

format/macho/macho.go
Akaame: Adopting the plural singular scheme.

format/macho/macho.go
Akaame: It think `d.FieldUTF8NullFixedLen("name", int(cmdsize)-int(offset))` is a better fit here. Taking that path.

format/macho/macho.go
```diff
@@ -0,0 +1,492 @@
...
+	"github.com/wader/fq/format/registry"
+	"github.com/wader/fq/pkg/bitio"
+	"github.com/wader/fq/pkg/decode"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+func init() {
+	registry.MustRegister(decode.Format{
+		Name:        format.MACHO,
+		Description: "Mach-O is the native executable format of binaries in OS X",
```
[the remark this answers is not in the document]
Akaame: Changed

format/macho/macho.go
Akaame: Changed

format/macho/macho.go
Akaame: Added.

Akaame:
Rookie question: do you use merge commits or do you rebase-force push?

pkg/interp/testdata/args.fqtest
```diff
@@ -91,6 +91,7 @@ id3v2                ID3v2 metadata
 ipv4_packet          Internet protocol v4 packet
 jpeg                 Joint Photographic Experts Group file
 json                 JSON
+macho                Mach-O is the native executable format of binaries in OS X
```
wader, line 94: `make actual` or `WRITE_ACTUAL=1 go test ./pkg/interp` should fix this

format/macho/macho.go
wader: Good, as long as the string looks user friendly, no nulls or extra spaces etc

format/macho/macho.go
wader: 👍

format/macho/macho.go
wader: Sorry i should have been more clear. What i meant was that magic seems to both indicate endian and number of bits? so it wold be nice if both of those var available as separate fields. One way could be something like thus:
```go
// should maybe assign map to some global var
d.FieldU32("magic", scalar.Hex, scalar.UToScalar{
	MH_MAGIC:    scalar.S{Description: "32-bit little endian"},
	MH_CIGAM:    scalar.S{Description: "32-bit big endian"},
	MH_MAGIC_64: scalar.S{Description: "64-bit little endian"},
	MH_CIGAM_64: scalar.S{Description: "64-bit big endian"},
})
d.FieldValueU("bits", uint64(archBits))
d.FieldValueStr("endian", ...)
```

If you would like to build some kind of shared table for this you can do something similar to this https://github.com/wader/fq/blob/master/format/elf/elf.go#L251 then you can use the map as normal and implement scalar.Mapper or have methods on the type that return mappers. But maybe overkill?

wader:
> No I need to make a huge table of all register values for that architecture and use it to decode `state` field, I am pushing that off for now as it is too much menial work 😄

Aha, yeah let's iterate on it.

> 🤦 why did I not think of this?? Thank you.

:) no problem, i did something much more complicated before i realize that was possible :)

> This is the general structure I followed if I got what you meant right.
> 
> e.g: `./fq '.load_command[9].dylib_command.compatibility_version' fq`

Good, i also meant i've started converting to lowercase and shorter symbols values etc, e.g: `.command=="symtab"` instead of `.command=="LC_SYMTAB"`.  The LC_ feels a bit redundant and i personally find it nicer to not have to write capitals so much also feels more jq/json-ish. What do you think?

> For test I will look at the tests for another format and adopt them. make lint fails for `cpuSubTypes` It will be fixed. Ran make doc.

👍 for elf i did it like this https://github.com/wader/fq/tree/master/format/elf/testdata i've noticed that it really helps a lot to have good tests when doing big refactors or working on a decoder, you will thank yourself :) also nice to have a script to instructions how to generate them if possible.

formats.md and format.svg will probably collide a lot, but is easy to just regenerate them.

wader:
> Rookie question: do you use merge commits or do you rebase-force push?

I personally prefer rebasing on master while working (and do fixups etc) instead of keep merging master into the PR branch. And then when done to a merge into master.

format/macho/macho.go
wader: 👍  no need to feel embarrassed :)

wader:
BTW when i tried the current macho branch it seemed to fail decoding, guess some change messed up something?

pkg/interp/testdata/args.fqtest
Akaame: Thanks for the help here

Akaame:
> BTW when i tried the current macho branch it seemed to fail decoding, guess some change messed up something?

Yes fixed it

> I personally prefer rebasing on master while working (and do fixups etc) instead of keep merging master into the PR branch. And then when done to a merge into master.

👍 

> .command=="symtab" instead of .command=="LC_SYMTAB"

👍

pkg/interp/testdata/args.fqtest
Akaame: Done thanks.

format/macho/macho.go
```diff
@@ -0,0 +1,574 @@
...
+}
+
+var endianNames = scalar.UToSymStr{
+	MH_MAGIC:    "little_endian",
+	MH_CIGAM:    "big_endian",
+	MH_MAGIC_64: "little_endian",
+	MH_CIGAM_64: "big_endian",
+}
+
+var cpuTypes = scalar.SToSymStr{
```
wader, line 46: Maybe should sync these with elf somehow? or at least shorter and lowercase?

format/macho/macho.go
```diff
@@ -0,0 +1,574 @@
...
+			archBits = 32
+		} else {
+			archBits = 64
+		}
+	} else {
+		d.Fatalf("Invalid magic field")
+	}
+
+	d.SeekAbs(0)
+	d.FieldStruct(fmt.Sprintf("mach_header_%d", archBits), func(d *decode.D) {
```
wader, line 319: Skip "mach_"? btw what do you think about just calling it "header" and then add archBits as a value field?

format/macho/macho.go
```diff
@@ -0,0 +1,574 @@
...
+					d.FieldU32("vmsize")
+					d.FieldU32("fileoff")
+					d.FieldU32("tfilesize")
+					d.FieldS32("initprot")
+					d.FieldS32("maxprot")
+					nsects = d.FieldU32("nsects")
+					d.FieldU32("flags") // TODO expand flags
+				})
+			} else {
+				d.FieldStruct("segment_command_64", func(d *decode.D) {
```
wader, line 359: Skip "_64"?

wader:
If you know where there will be opcodes at some point it would be nice with a comment or something. Think i will try add opcode decoders for various ISA:s at some point.

Decoder works fine now, looks good! noticed that some string field have nulls, would be nice to strip i think? 

Let me know when you feel the decoder covers enough things to merge initial support.

Akaame:
> If you know where there will be opcodes at some point it would be nice with a comment or something. Think i will try add opcode decoders for various ISA:s at some point.

`segment_command` and `segment_command_64` have `__TEXT` subsections which will have opcodes. Let us decide on a special mark so that when you add the support we can directly integrate them like `// OPCODE_DECODER`

> Decoder works fine now, looks good! noticed that some string field have nulls, would be nice to strip i think?

Can you show which ones those are. I am using UTF8NullFixedLen which stopped at null bytes.

> Let me know when you feel the decoder covers enough things to merge initial support.

- Naming changes you mentioned.
- Flag expansions
- Working on THREADSTATE parsing
- If possible FAT binaries would be good (bunch of macho files glued together with binaries for different ISAs that requires an extra header parse)

wader:
> > If you know where there will be opcodes at some point it would be nice with a comment or something. Think i will try add opcode decoders for various ISA:s at some point.
> 
> `segment_command` and `segment_command_64` have `__TEXT` subsections which will have opcodes. Let us decide on a special mark so that when you add the support we can directly integrate them like `// OPCODE_DECODER`

👍  yeap does not have to be very formal just a short todo reminder

> > Decoder works fine now, looks good! noticed that some string field have nulls, would be nice to strip i think?
> 
> Can you show which ones those are. I am using UTF8NullFixedLen which stopped at null bytes.

Ah it's `segname` and `sectname`. You can do something like `go run fq.go 'vgrep("\x00")' fq` to look for strings with nulls. Might need to rebase on master, i did some fixes to grep related to this some days ago.

> > Let me know when you feel the decoder covers enough things to merge initial support.
> 
> * Naming changes you mentioned.
> * Flag expansions
> * Working on THREADSTATE parsing
> * If possible FAT binaries would be good (bunch of macho files glued together with binaries for different ISAs that requires an extra header parse)

👍  was going to ask about FAT-binaries, got a bit curious how they work

format/macho/macho.go
Akaame: I think it should be tackled in some other task

format/macho/macho.go
Akaame: Done.

format/macho/macho.go
Akaame: Done similarly

Akaame:
> Yeap does not have to be very formal just a short todo reminder

Added the reminder

> Ah it's segname and sectname. You can do something like go run fq.go 'vgrep("\x00")' fq to look for strings with nulls. Might need to rebase on master, i did some fixes to grep related to this some days ago.

Was an overlook from my side. Fixed

format/macho/macho.go
wader: 👍

format/macho/testdata/darwin_amd64/libbbb.a.fqtest
```diff
@@ -0,0 +1,31 @@
+     |00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f|0123456789abcdef|.{}: darwin_amd64/libbbb.a (ar) 0x0-0x3b7.7 (952)
```
wader, line 1: Any idea what format is used in archive on macOS? On linux the .o files inside archives turned out to be ELF, i for some reason assumed it used something else

wader:
Nice with test files! 

BTW maybe should double check what static builds on macOS actually does, there seems to be some conflicting messages if it actual makes a difference.

Akaame:
There is actually a separate path of parse for .a files. That is why we are seeing those errors during testing. Tests are broken atm anyway.

I will fix the tests(leaving .a out), rebase and start working on AR(.a) and FAT binaries.

format/macho/testdata/darwin_amd64/libbbb.a.fqtest
Akaame: So the file can be a plain macho file (as we have been parsing until now) a fat file or an ar file.

Currently what I am implement on my local

Check if it is plain
If not check if it is FAT,
If not try AR file parsing,
And if that also fails say that this file has an invalid magic identifier.

wader:
Hey, i did some cleanup/sorting related to list of formats, it probably caused some collision but after resolve them i think they will be less likely to happen

format/macho/macho.go
```diff
@@ -559,7 +586,71 @@ func machoDecode(d *decode.D, in interface{}) interface{} {
...
+
+func arParse(d *decode.D) {
+	// Go to start of the file again
+	d.SeekAbs(0)
+	d.FieldStruct("ar_header", func(d *decode.D) {
+		name := d.FieldUTF8NullFixedLen("name", 16)
+		d.FieldUTF8NullFixedLen("date", 12)
+		d.FieldUTF8NullFixedLen("uid", 6)
+		d.FieldUTF8NullFixedLen("gid", 6)
+		d.FieldUTF8NullFixedLen("mode", 6)               // parse_octal ?? TODO
```
wader, line 626: Can do something like https://github.com/wader/fq/blob/master/format/ar/ar.go#L31 .. wonder some more generic sym mapper should be added to the api for this, seems common.

format/macho/macho.go
```diff
@@ -559,7 +586,71 @@ func machoDecode(d *decode.D, in interface{}) interface{} {
...
+			})
+		})
+		for i := 0; uint64(i) < narchs; i++ {
+			// parse ofiles
+			ofileDecode(d)
+		}
+	})
+}
+
+func arParse(d *decode.D) {
```
wader, line 618: Could the existing ar decoder be used for this or is it too macho-specific? It looks like ar but without the `"!<arch>\n"` header? to reuse it i guess it needs some option to skip header and something to hint it how to handle files with certain names to handle SYMDEF and nested ar? maybe too messy

wader: Ah there is a header, just noticed the ARMAG thing

Akaame: I totally forgot that we should have an existing ar decoder. Let me actually delete this and re-use the already tested one.

Akaame:
I feel like this PR has been dragging on too much and I can only contribute during weekends so it will continue like this. Can we at some point merge this (with its todos and missing parts) and work on the improvements on some other branch or is the pace fine by you?

format/macho/macho.go
Akaame: I saw this now and updated accordingly.

wader:
> I feel like this PR has been dragging on too much and I can only contribute during weekends so it will continue like this. Can we at some point merge this (with its todos and missing parts) and work on the improvements on some other branch or is the pace fine by you?

I'm fine with slow pace, happy for any help i can get :) but it's usually not a good idea to have long running branches, so i agree maybe good idea focus on tidy up the current code and add nice TODOs and get something merged?

Akaame:
> so i agree maybe good idea focus on tidy up the current code and add nice TODOs and get something merged?

okay I will work on it this weekend so that at least we have something somewhat tangible on upstream.

format/macho/macho.go
Akaame: deleted the code and re-used ar parser from ar format.

That required main decoder method to be exported as `DecodeAr`. I hope it is fine.

wader: You mean exported https://github.com/wader/fq/blob/master/format/ar/ar.go#L27? I try to keep the format implementations decoupled from each other to not cause import cycles and also to force them to have a somewhat clear APIs between each other. So i would prefer if you used the `format.Dependencies` to register that the format needs `fornmat.AR` and then you can use `d.Format*(...)` to actually do the decoding.

Akaame: oh okay. I was not aware of this usage. Will look into its usages.

wader: 👍  So if you want communicate something to the ar decoder you would create a `format.ArIn` struct and somehow model what you want do, say maybe in this case a map saying that some filenames are suppose to be parser a certain way or that it should try to do nested ar-archives decode etc, but it can be a bit tricky to modell these things nicely. You can also return data by creating a `format.ArOut` struct and return it. But all of this  requires that you do type asserts and such things. 

I have some future plans to allow a user to pass these format specific options directly from jq, ex: `mp4({skip_sample_decode: true})`.

format/macho/macho.go
wader: Did a PR to factor this out https://github.com/wader/fq/pull/111

format/macho/testdata/darwin_amd64/libbbb.a.fqtest
```diff
@@ -0,0 +1,33 @@
+$ fq -d macho v libbbb.a
```
wader, line 1: Could there be bytes before the ar archive starts? if not i wonder if both the ar and macho decoder will overlap now. Maybe is would make sense to add macho heuristics to the ar decoder instead? maybe even have SYMDEF as a sub format in some group and probe for in the ar decoder? But if we don't know yet maybe better to remove macho ar things for now and add it in separate PR?

Akaame: We can leave archive parsing out directly. I think it is cross platform. We can add it later on if there is demand. But I do not think so.

wader: 👍

Akaame:
TODOs are resolved. The only big thing remaining rn is test cases for FAT binaries and if possible ARM64. Let us handle that in another issue. I will fix the errors in CI and then the thing is ready to be reviewed wholly.

wader:
@Akaame 👍  looks good overall i think, i will give it a closer review during the week and also take it for a spin!

wader:
I did a `go run fq.go -d macho dv /bin/*` and go some crashes about type asserts in the fat decoder. I guess `d.FieldU32("cputype", cpuTypes)` should be `S32` and also cpusubtype? wonder also if fat magic is only 32 bit?

Was the plan to skip fat for this PR? at least it would be nice to not cause crashes and instead crash in a controlled away using `d.Fatal` etc

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+//nolint:revive
+const (
+	MH_MAGIC    = 0xfeedface
+	MH_CIGAM    = 0xcefaedfe
+	MH_MAGIC_64 = 0xfeedfacf
+	MH_CIGAM_64 = 0xcffaedfe
+	FAT_MAGIC   = 0xcafe_babe
+	FAT_CIGAM   = 0xbeba_feca
+	ARMAG       = "!<arch>\n"
+	AR_EFMT1    = "#1/"
```
wader, line 34: Not used now?

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+}
+
+//nolint:revive
+const (
+	MH_MAGIC    = 0xfeedface
+	MH_CIGAM    = 0xcefaedfe
+	MH_MAGIC_64 = 0xfeedfacf
+	MH_CIGAM_64 = 0xcffaedfe
+	FAT_MAGIC   = 0xcafe_babe
+	FAT_CIGAM   = 0xbeba_feca
```
wader, line 32: Maybe "_"-group the other value also?

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+				d.FieldU32("reserved6", d.BitBufIsZero())
+			} else {
+				d.FieldU64("init_address", scalar.Hex)
+				d.FieldU64("init_module")
+				d.FieldU64("reserved1", d.BitBufIsZero())
+				d.FieldU64("reserved2", d.BitBufIsZero())
+				d.FieldU64("reserved3", d.BitBufIsZero())
+				d.FieldU64("reserved4", d.BitBufIsZero())
+				d.FieldU64("reserved5", d.BitBufIsZero())
+				d.FieldU64("reserved6", d.BitBufIsZero())
```
wader, line 573: BitBufIsZero only works if the actual value is a buffer, so this is probably a nop. Maybe it should even panic if used with wrong type

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+		}, func(d *decode.D) {
+			// parse FatArch
+			d.FieldStruct("fat_arch", func(d *decode.D) {
+				// beware cputype and cpusubtype changes from ofile header to fat header
+				cpuType := d.FieldU32("cputype", cpuTypes)
+				d.FieldU32("cpusubtype", cpuSubTypes[int64(cpuType)])
+				d.FieldU32("offset")
+				d.FieldU32("size")
+				d.FieldU32("align")
+			})
```
wader, line 691: narchsIdx++ missing?

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+	})
+	d.FieldU32("sp")
+	d.FieldU32("lr")
+	d.FieldU32("pc")
+	d.FieldU32("cpsr")
+}
+
+func threadStateARM64Decode(d *decode.D) {
+	rIdx := 0
+	d.FieldStructArrayLoop("r", "r", func() bool {
```
wader, line 783: Some of the FieldStructArrayLoop seems to be missing increment of the loop variable?

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+			d.FieldStruct("source_version_tag", func(d *decode.D) {
+				d.FieldU64("tag")
+			})
+		case LC_LINKER_OPTION:
+			d.FieldStruct("linker_option", func(d *decode.D) {
+				count := d.FieldU32("count")
+				d.FieldUTF8NullFixedLen("option", int(count))
+			})
+		case LC_ENCRYPTION_INFO, LC_ENCRYPTION_INFO_64:
+			d.FieldStruct(fmt.Sprintf("encryption_info_%d", archBits), func(d *decode.D) {
```
wader, line 652: Skip %d in struct name? will always be in the header?

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+						d.FieldU64("address", scalar.Hex)
+						d.FieldU64("size")
+					}
+					d.FieldU32("offset")
+					d.FieldU32("align")
+					d.FieldU32("reloff")
+					d.FieldU32("nreloc")
+					// get section type
+					sectionType := d.U32()
+					sectionTypeMasked := sectionType & 0xFF
```
wader, line 496: I like to keep the casing consistent, so 0xff would be good

format/macho/macho.go
```diff
@@ -0,0 +1,835 @@
...
+		d.FieldValueS("arch_bits", int64(archBits))
+		magic := d.FieldU32("magic", scalar.Hex, magicSymMapper)
+		d.FieldValueU("bits", uint64(archBits))
+		d.FieldValueStr("endian", endianNames[magic])
+		cpuType = d.FieldS32("cputype", cpuTypes)
+		d.FieldS32("cpusubtype", cpuSubTypes[cpuType])
+		d.FieldU32("filetype", fileTypes)
+		ncmds = d.FieldU32("ncdms")
+		d.FieldU32("sizeofncdms")
+		d.FieldStruct("flags", parseFlags(machHeaderFlags))
```
wader, line 437: I wonder if the two uses for  parseFlags could be replaced with a decodeMachineHeader and decodeSegmentFlag functions that uses a bunch of d.FieldBool("...") calls, something like https://github.com/wader/fq/blob/master/format/elf/elf.go#L773? that way all flags will always be available and also "addressable".

Akaame:
> I did a go run fq.go -d macho dv /bin/* and go some crashes about type asserts in the fat decoder. I guess d.FieldU32("cputype", cpuTypes) should be S32 and also cpusubtype? wonder also if fat magic is only 32 bit?

FAT magic is U32 and cpuType and subType is also U32. This was one of the weird changes I encountered between vanilla OFile and FAT.

https://go.dev/src/debug/macho/fat.go

I did a run on my wife's M1 and reproduced the crash you had.

So we are having trouble now because for FAT cpu types are unsigned but SymMapper itself is signed as OFile cpu types are signed. What is a possible way out of this?

wader:
> > I did a go run fq.go -d macho dv /bin/* and go some crashes about type asserts in the fat decoder. I guess d.FieldU32("cputype", cpuTypes) should be S32 and also cpusubtype? wonder also if fat magic is only 32 bit?
> 
> FAT magic is U32 and cpuType and subType is also U32. This was one of the weird changes I encountered between vanilla OFile and FAT.
> 
> https://go.dev/src/debug/macho/fat.go
> 
> I did a run on my wife's M1 and reproduced the crash you had.

👍 

If you have go tip installed you can also use `make fuzz` to run the fuzzer, but currently it will always fuzz all formats, it's quite effective at findings hangs and crashes.

> So we are having trouble now because for FAT cpu types are unsigned but SymMapper itself is signed as OFile cpu types are signed. What is a possible way out of this?

Aha now i understand the comment in the code, cpuTypes/cpuSubTypes is used for both but reads different integer types. That is an interesting challenge. Are the lists exactly the same except the for -1 (any/multiple)? I see some possible solutions:

- Have separate maps
- Implement own ScalarMapper type that knows that actual type might be S or U
- Read both as S32, it seems like the other values are all same for S32 or U32 none has bit 31 set?

Now i got curious what CPU type any is used for :) shared resources somehow?

Akaame:
> Read both as S32, it seems like the other values are all same for S32 or U32 none has bit 31 set?

So I have taken a look at Rust's macho parser and saw this.

```pub const CPU_TYPE_ANY: u32 = !0; // 0_xff_fff_fffu32```

is it safe to assume two's complement? If it is we can just adopt to this and avert the trouble. If not, I would like the select the:

> Implement own ScalarMapper type that knows that actual type might be S or U

**Comment by wader** [comment]
> > Read both as S32, it seems like the other values are all same for S32 or U32 none has bit 31 set?
> 
> So I have taken a look at Rust's macho parser and saw this.
> 
> `pub const CPU_TYPE_ANY: u32 = !0; // 0_xff_fff_fffu32`
> 
> is it safe to assume two's complement? If it is we can just adopt to this and avert the trouble. If not, I would like the select the:
> 
> > Implement own ScalarMapper type that knows that actual type might be S or U
> 
> Edit: On second thought assuming two's complement is fine. I would propose to go with it and convert everything to U.

Yeap think so too, try read both as U32 and map 0xff_ff_ff_ff to any

**Comment by Akaame** [comment]
Tested on M1 and X86-64. @wader please pull and confirm.

File: format/macho/macho.go [hunk 26]
```diff
@@ -0,0 +1,851 @@
...
+			})
+		case LC_IDFVMLIB, LC_LOADFVMLIB:
+			d.FieldStruct("fvmlib", func(d *decode.D) {
+				offset := d.FieldU32("offset")
+				d.FieldU32("minor_version")
+				d.FieldU32("header_addr", scalar.Hex)
+				d.FieldUTF8NullFixedLen("name", int(cmdsize)-int(offset))
+			})
+		default:
+			if _, ok := loadCommands[cmd]; !ok {
```
wader, line 674: Is there any command not handled in a cases above that are in loadCommands? if so i think they will be skipped and get no unknown field? btw i've used "unknown" fields for gaps that fq finds automatically so maybe good to use some other word like "data"?
If something is truly unknown the idea is for a decoder not even add any fields for it and let fq add gap fields instead

format/macho/macho.go
```diff
@@ -0,0 +1,851 @@
...
+		if archBits == 64 {
+			d.FieldRawLen("reserved", 4*8, d.BitBufIsZero())
+		}
+	})
+	ncmdsIdx := 0
+	d.FieldStructArrayLoop("load_commands", "load_command", func() bool {
+		return ncmdsIdx < int(ncmds)
+	}, func(d *decode.D) {
+		cmd := d.FieldU32("cmd", loadCommands)
+		cmdsize := d.FieldU32("cmdsize")
```
wader, line 456: I wonder if you could use `d.FramedFn(...)` here if no decoder inside of it needs to dead outside the framed range. Then for example `d.FieldRawLen("unknown", int64((cmdsize-8)*8))` at the end can be `d.FieldRawLen("unknown", d.BitsLeft())`

wader:
Yeap no crashes for me either. some thoughts:
- Change `parseFlags` into just list of `d.Fieldbool(...)` and leave out the "flags" field. That way each bit will have range and be addrssable. I usually try to avid using `d.Field*Value` and use it only for things that have no resonable range in the file.
- Possible to generate ARM tests?
- Most macho binaries on macOS seems be FAT. Possible to fix that before merge? thinking maybe misleading to announce macho support if i do a release before FAT support i done.

Akaame:
> Most macho binaries on macOS seems be FAT. Possible to fix that before merge? thinking maybe misleading to announce macho support if i do a release before FAT support i done.

That is because you have a M1 chip I am guessing.

> Possible to generate ARM tests?

It is definitely possible. I did not have time for cross compiling to ARM for test generation (and generating fat binaries for fat tests).

> Change parseFlags into just list of d.Fieldbool(...) and leave out the "flags" field. That way each bit will have range and be addrssable. I usually try to avid using d.Field*Value and use it only for things that have no resonable range in the file.

👍

wader:
> That is because you have an M1 chip I am guessing.

I'm sadly on an older x86 macbook. file says this about /bin/ls:

```
$ file /bin/ls
/bin/ls: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64e:Mach-O 64-bit executable arm64e]
/bin/ls (for architecture x86_64):	Mach-O 64-bit executable x86_64
/bin/ls (for architecture arm64e):	Mach-O 64-bit executable arm64e
```

> It is definitely possible. I did not have time for cross compiling to ARM for test generation (and generating fat binaries for fat tests). I am also waiting for an M1 delivery in the meantime so have postponed this task until now. Will hit you up with the new developments this weekend

How is it generally done? build x86 and arm marcho binaries separetly and then use some tool to merge them into a fat? Have some vague memory of doing this with x86 and powerpc binaries many years ago

I can probably do some research into it also. Maybe clang is nicer fort this? could use go cross build? :) but maybe will be binaries that are not very representative?

I should get a new laptop also, keys are getting loose :)

wader:
This seems to be the steps:
```
$ echo 'void main() {}' > test.c
$ clang test.c -o test_x86_64 -target x86_64-apple-macos10.12
$ clang test.c -o test_arm64 -target arm64-apple-macos11
$ lipo -create -output test_fat test_x86_64 test_arm64
$ file test_fat
test_fat: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
test_fat (for architecture x86_64):	Mach-O 64-bit executable x86_64
test_fat (for architecture arm64):	Mach-O 64-bit executable arm64
```

Akaame:
Found the stupid error I made for fat binaries. Will push it along with the generated tests tomorrow. Thanks for the help again.

wader:
👍  No problem

wader:
This was quite useful when i stress tested the elf decoder:
```
fq -d elf 'select(._error and (tobytes | test("^#!") | not))' /path/elfs/*
```
It will only show files that failed and also filter out shell scripts. But beware that it might use a lot of fd:s as fq is a bit bad at the moment to cleanup those.

Akaame:
@wader

FAT parsing bugs are resolved.
ARM and FAT test cases and auto generation is added.
Unknown LC codes are handled with SeekRel so we get "unknown" handling by decoder itself.
Header, Segment and Section Flags are parsed with `FieldBool`

format/macho/macho.go
```diff
@@ -0,0 +1,868 @@
...
+		}
+		ncmdsIdx++
+	})
+}
+
+func fatParse(d *decode.D) {
+	// Go to start of the file again
+	d.SeekAbs(0)
+	var narchs uint64
+	var ofile_offsets []uint64
```
wader, line 638: `ofileOffsets`? wonder if some other vars should be snakeCased also like `nArchs`?

format/macho/macho.go
wader: Thinking it might be better to add this as "data" etc instead of seeking paste by it. Sorry maybe should have explained more in details what the idea of auto "unknown" fields are usually used for. For example in this case we know the command number and the size so we could add "data" field saying these bits belong to the command at least. But yes it's a bit of balance sometimes what to skip , what to add as something close to "unknown".

format/macho/macho.go
```diff
@@ -0,0 +1,868 @@
...
+func parseSectionFlags(d *decode.D) {
+	d.FieldBool("S_ATTR_PURE_INSTRUCTIONS")
+	d.FieldBool("S_ATTR_NO_TOC")
+	d.FieldBool("S_ATTR_STRIP_STATIC_SYMS")
+	d.FieldBool("S_ATTR_NO_DEAD_STRIP")
+
+	d.FieldBool("S_ATTR_LIVE_SUPPORT")
+	d.FieldBool("S_ATTR_SELF_MODIFYING_CODE")
+	d.FieldBool("S_ATTR_DEBUG")
+	d.RawLen(1)
```
wader, line 728: Maybe replace `RawLen`:s with one `d.FieldRawLen("unused0", 14)` etc? and the next with `d.FieldRawLen("unused0", 8)`? same reasoning as for unknown cmd bits,  bits as not totally unknown. These flags  bits are unused (padding?) or reserved i guess?

format/macho/macho.go
```diff
@@ -0,0 +1,868 @@
...
+						d.FieldU64("size")
+					}
+					d.FieldU32("offset")
+					d.FieldU32("align")
+					d.FieldU32("reloff")
+					d.FieldU32("nreloc")
+					// get section type
+					sectionType := d.U32()
+					sectionTypeMasked := sectionType & 0xff
+					d.FieldValueStr("type", sectionTypes[sectionTypeMasked])
```
wader, line 457: Thinking maybe `sectionType := d.U32() .... d. FieldValueStr()` can be replaced by something like:
```go
d.SeekRel(24)
d.FieldU8("type", sectionTypes)
d.SeekRel(-32)
d.FieldStruct("flags", parseSectionFlags)
...
```
or make `parseSectionFlags` only read 24 bits and have `d.FieldU8("type", sectionTypes)` after?

Is section type part of the flags in the spec?

format/macho/macho.go
```diff
@@ -0,0 +1,868 @@
...
+		d.FieldStruct("flags", parseMachHeaderFlags)
+		if archBits == 64 {
+			d.FieldRawLen("reserved", 4*8, d.BitBufIsZero())
+		}
+	})
+	ncmdsIdx := 0
+	d.FieldStructArrayLoop("load_commands", "load_command", func() bool {
+		return ncmdsIdx < int(ncmds)
+	}, func(d *decode.D) {
+		cmd := d.FieldU32("cmd", loadCommands)
```
wader, line 406: Maybe nice to add `scalar.Hex` if cmds are in hex in the spec

wader:
@Akaame Hey, hope all is fine! just want to check if there is something i can help out with

format/macho/macho.go
Akaame: `ofileOffsets` fix is made sorry I was probably transitioning from work there and used snake_case 😆 
For `nfoobar` variable I would not go with camel case really

Akaame:
@wader thanks for asking. I was busy with some other stuff. Back in action now.

format/macho/macho.go
Akaame: The flag bits are reserved as they do not correspond to anything as of now. I would not go with making them `FieldRawLen` as we have no reason to make them addressable as far as I see. The reason they are grouped by 4-bit groups is that I found it more reasonable to keep track of during impl. Will change this for this and other methods.

Akaame: Is the reasoning here with `FieldRawLen` to make these fields, we have skipped over, not be marked as `unknown_`?

Akaame: I see that is the case as you have mentioned the empty slots withing decoder will be marked as unknown and clutter the result. I would still go with using the keyword `reserved` for `FieldRawLen`

Akaame:
@wader can you please re-test the latest version of this branch?

format/macho/macho.go
wader: Yeap "reserved" is good here i think. Maybe there is better word then "unknown" for bits not claimed by a decoder? hard to come up with good termonology.

Maybe was unclear with addressable, didn't mean making each reserved bit a separetly, i think it's good as it is now. Now someone could make reasonable queries selecting on reserved bits and a the same time it decode tree looks nice i think.

format/macho/macho.go
wader: 😃  know the feeling, can be a bit confusing in this project, jq uses snake_cake

format/macho/macho.go
wader: What do you think about this? maybe most of the unknown i saw as actually just commands that are handled but the offset/size fields are not used at the moment? for ex LC_CODE_SIGNATURE and some others (hard to do generic?) there is offset and size that could be added a as raw "data". Maybe for later?

wader:
Think we're almost ready for merge, do you have something more you want to add or fix?

BTW if you wan't to add macho specific documentation you can add a `macho.md` file and it will be included into `formats.md` when you run `make doc`. See ex `asn1_ber.md` if you want some inspiration.

wader:
> @wader can you please re-test the latest version of this branch?

Sure, no crashes for the binaries on my macbook and also ran the fuzzer for a while without issues 👍

wader:
> @wader thanks for asking. I was busy with some other stuff. Back in action now.

👍  good to hear and no worries!

Akaame:
@wader added some basic docs

format/macho/macho.md
```diff
@@ -0,0 +1,18 @@
+# MACHO Format
+
+Decodes vanilla and FAT Macho binaries
+
+## Examples
+
+```
+fq . -d macho fq
```
wader, line 8: Maybe add that `-d` is usually not needed?

wader:
> @wader added some basic docs

👍  let know if your think it's ready for merge

Akaame:
@wader I think we can continue with the merge. It is better to showcase this and get some feedback from the users and iterate if they have pain points.

wader:
Thanks very much for this, good work!

PR #69, opened 2022-01-10
msgpack: Add decoder

### LINKED ISSUE #25: Support for MessagePack (msgpack)
## Description

`fq` looks like a great tool! Checking out the supported formats, it doesn't appear that https://msgpack.org/ is supported at the moment.

> MessagePack is an efficient binary serialization format. It lets you exchange data among multiple languages like JSON.

Is this format something that would be considered for this project? It would be great to be able to inspect and filter against msgpack data.

wader:
Fixes #25


bflad (approved):
From a functionality standpoint, this looks great to me. Built locally and given a saved msgpack data file, I was able to decode using `d`, `tovalue`, and `msgpack_torepr` with helpful output. Thank you, @wader!

wader:
👍  Great, thanks for testing. I hope to make the `*_torepr` things simpler in the future, currently thinking there should be a `torepr` function that automatically converts to a representation if the format supports it. Then the usage would be `fq torepr file` or maybe even some command line flag.

wader:
And let me know if something seems to behave strange, i'm not very familiar with msgpack.

PR #73, opened 2022-01-11
doc: Add fq usage guide solving part of FRA pcap-challenge-2021-09

wader:


Doctor-love:
Personally, I think this is pretty awesome! Shows the power of fq and the jq language - guess most people don't know much more than basic queries.

I say more!

Doctor-love:
Mayhaps use generated SVGs in examples to show off the highlighting and making it easier to follow?

Doctor-love:
> Mayhaps use generated SVGs in examples to show off the highlighting and making it easier to follow?

But harder to copy-paste... hmm.

wader:
Ah good idea. Maybe ok to now be able to copy/paste from them? wonder how it would look if only the output as SVG and let the command above be plain old markdown?

Was it ok to follow? i think it might go a bit too fast at the end currently, will try break that up a bit.

Doctor-love:
I don't think it would hurt to explain the syntax and functions more, but this is a good start - a few TODOs left to fix.

Doctor-love:
Both the ideas you mention sounds interesting - especially broken files. I would guess that the biggest hurdle for people is to even know where to start with fq, considering the vast amount of jq functions and similar - perhaps try to structure a guide that covers the basics and becomes progressively more advanced, the actual problem that is solved in the guide could be more or less synthetic.

wader:
So maybe a pure jq guide would be good? could also maybe include some CLI stuff?

Doctor-love:
> So maybe a pure jq guide would be good? could also maybe include some CLI stuff?

jq from fq's perspective, yeah :-)

Doctor-love:
Small tweak: perhaps a git mv to fix spelling in filename of "doc/guides/fra-pcap-**challange**-2021-09/README.md"

Issue #20, opened 2021-12-23
Feature idea: support ASN.1 (DER/BER)


gjvnq:
Hey, it would be pretty cool with fq supported decoding binary¹ [ASN.1](https://en.wikipedia.org/wiki/ASN.1#Language_support) (Abstract Syntax Notation One) as it is used in various communication protocols.

¹ ASN.1 has many encoding options but pretty much everything that is binary encoded seems to be either BER or DER.

The main problem I see is that the field names are not included in the actual files, so either the user would have to point to an ASN.1 grammar/spec file or the output would be "nameless".

wader:
Hi, yes that is a good idea. It would also require figuring out a nice way of passing in a grammar file to the decoder.
If you want to work on this i would recommend having a look at the protobutf decoder https://github.com/wader/fq/blob/master/format/protobuf/protobuf.go. It does not support passing in schemas but other decoders can pass it something schema-like internally to improve the decode output.

gjvnq:
Thanks. I will take a look if I have time in January.

wader:
👍  There are some other issues #25 #24 that will share som issues with this like how to provide an external description file and how to model the decoded value tree.

wader:
Hey, some update. I have a basic BER decoder working now, also working on adding support for arbitrary large integer to the decoder which will be useful for ASN.1. and make it more useful for certificates and crypto related things i think. Hope i can push something in the next few days.

wader:
Hey, now there is basic ASN1 BER support in master and also some kind of `torepr`  support.

To make it a bit easer to work with PEM there is also `frompem`/`topem` now.

Any feedback what could be done differently or bugs are much appreciated.

Example reading PEM file, decode it and display it as decode tree, JSON and torepr JSON:
```
$ fq -o line_bytes=10 -d raw '(. | tobytes | tostring | println), (frompem | asn1_ber | d, tovalue, torepr)' format/asn1/testdata/test.pem
-----BEGIN PUBLIC KEY-----
MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDMYfnvWtC8Id5bPKae5yXSxQTt
+Zpul6AnnZWfI2TtIarvjHBFUtXRo96y7hoL4VWOPKGCsRqMFDkrbeUjRrx8iL91
4/srnyf6sh9c8Zk04xEOpK1ypvBz+Ks4uZObtjnnitf0NBGdjMKxveTq+VE7BWUI
[...]

PR #92, opened 2022-01-20
asn1_ber: Add decoder

wader:
Related to #20


wader:
Very basic and many things still broken, bit it's a start. I think i will probably skip grammar things for now, would have to do a quite complicated parser for that.

Doctor-love:
Let us know if you need mental and emotional support after this effort

wader:
To decode some asn1 types correctly i need some kind of proper bit buffer, also need that to fix proper bits -> bytes "iolist" serialization

PR #108, opened 2022-01-26
Polish of problem template and clarifying questions

Doctor-love:


.github/ISSUE_TEMPLATE/issue.md
```diff
@@ -28,6 +35,6 @@ To get more information about what is going on you can try to force a format ask
 output using "fq -d <format> v file"
```
wader, line 35: Noticed a typo "force a format __and__ ask for"

.github/ISSUE_TEMPLATE/issue.md
```diff
@@ -14,7 +14,14 @@ $ fq -v
 
 </pre>
 
-### Does this issue reproduce with the latest release or master?
+### How was fq installed?
+
+<!--
+Please describe how/from what source fq was installed, for example using
+FreeBSD ports, Alpine experimental repository or manually built from source
```
wader, line 21: With this maybe can remove the "What operating system and processor architecture are you using?" question below? think it's good to not ask too many questions :)

Doctor-love: That's very true - what do you think is most relevant? Considering the recent reports, my guess is that installation method is a better guide during triage.

Thinking about it, wouldn't something like this be neat?
```
-- ~/devel/fq: docker run --rm fq:latest --version
0.0.4 (Linux x86_64)
```

That way, by asking for the version we don't need to query them for OS/arch.

wader: Ah yeah good idea, will add that

PR #112, opened 2022-01-28
interp: Cleanup display aliases, now: d, da, dd, dv, ddv

wader:
Think it makes sense to have them all start with d.
Also f is often used as function argument name.


wader:
@Akaame @xentripetal With this PR `verbose` function will be replaced with just `dv`, so non updated fqtest:s will break. I think it makes sense to call all display aliases `d<something>`, the current `v`, `verbose`, `f` and `full` feel a bit confusing. Let me know if you think it's a bad idea.

PR #123, opened 2022-02-04
bitio,decode: Refactor bitio usage and make buffer slicing more correct

wader:
Remove bitio.Buffer layer. bitio.Buffer was a kitchen sink layer with helpers
now it's just a buffer and most functions have been moved to decode instead.

bitio package now only have primitive types and functions simialar to standard
library io and bytes packages.

Make nearly eveything internally use bitio.Bit* interfaces so that slicing work
correctly this will also make it possible to start experimenting with more
complicated silcing helpers, ex things like:
breplace(.header.bitrate; 123) to get a new buffer with bitrate changed.


wader:
Once this is merged we can finally start experimenting with "modifying" data in an easier way, it will allow for slicing buffers at bit level based on decode values and ranges etc and then concat things together and get serialized bit or bytes back.

For example this works now:
```
$ fq.go -n '[0,1,1,0,0,1,1,0,0,1,1,1,0,0,0,1 | tobits] | tobytes | tostring'
"fq"
```

@Akaame @xentripetal This refactor is quite big but should not affect decode API that much. I hope in worst case for you it might mean changing some int to int64 in a decode API call.

PR #133, opened 2022-02-08
binary,decode,doc: Rename buffer to binary and add some documentation

wader:
Rename buffer to binary. Still some work left what to call buffer/binary in decode code.
Document decode value and binary type
Fix proper unit padding for tobytes and add still undocumenated extra padding argument.
Add some additional binary tests


wader:
@Doctor-love if you have time some review of the new documentation would be nice

PR #138, opened 2022-02-10
doc: Add more license details

wader:
Related to #137


wader:
@thushan should be enough?

Issue #98, opened 2022-01-24
Build and publish Snap image for releases


Doctor-love:
["Snap"](https://en.wikipedia.org/wiki/Snap_(package_manager)) is a popular package format that enables Linux users to install and run applications cross-distro.
In order to help potential users discover and use fq, it would be neat to automatically build and publish a snap during the release process.

fq utilizes [GoReleaser](https://github.com/goreleaser/goreleaser) in [the "release" workflow](https://github.com/wader/fq/blob/master/.github/workflows/release.yml), which [according to the document supports building and publishing snaps](https://goreleaser.com/customization/snapcraft/).

I don't have any experience packaging snaps, so help would be greatly appreciated.

Doctor-love:
How is confinement handled in Snaps? fq should be very easy to restrict (syscalls, network access, etc), except for which file paths it needs to access for probing.

wader:
Yeap i think fq has very basic needs, read filsystem, read/write stdout and stderr maybe some ioctls for readline. All interaction with the os (exceptions for some profiling and fuzz things) are done thru the `stdOS` type https://github.com/wader/fq/blob/master/pkg/cli/cli.go#L36

Maybe first step i get a snap config with good confinement  working and then look into goreleaser things?

I can do -rc tag releases once we have something to try

Akaame:
```yaml
name: fq
summary: jq for binary formats 
description: |
    fq is inspired by the well known jq tool and language and allows you to work with binary formats the same way you would using jq. 
    In addition it can also present data similar to a hex viewer, transform, slice and concatenate binary data, supports nested formats and has an interactive REPL with auto-completion.
version: git
grade: stable
base: core20

confinement: classic # or strict

apps:
  fq:
    command: bin/fq
    plugs:
      - home # should be enough
parts:
  fq:
    plugin: go
    source-type: git
    source: https://github.com/wader/fq
```

Do not have a linux system to test this on but should not something like this be enough for a basic golang based IO program?

wader:
Thanks, i guess one maybe can test via docker or emulated if on some non-snap:able platform? what is needed more, some kind of packaging and publish? @Doctor-love any ideas?

@Akaame BTW I noticed your name didn't show up as contributor on github for fq, maybe git email configured differently somehow?

@Akaame Second BTW if your into binary tools there is a conference going on this weekend https://binary-tools.net/summit.html. I gave a talk about fq yesterday that was recorded so hopefully will show up somewhere soon.

[...]

PR #189, opened 2022-03-09
fq: Update to version 0.0.6

wader:


wader:
Still WIP, let me know if there is something someone would like to add

wader:
@dundee @siraben @thushan @danfe @mpanczyk @orhun Hello, 0.0.6 have been published 🎉 

Have been mostly focused on fixing and cleaning up the REPL and CLI usage. But @Akaame macho decoder was finally merged. Next release will probably be more focused on formats and decoder things.

See [CHANGES.md](https://github.com/wader/fq/blob/master/CHANGES.md) for change and also the project description have been polished a bit if you want to update it in some package description etc.

herbygillot:
Thanks for the heads-up, @wader. `fq` has already been updated in MacPorts: https://ports.macports.org/port/fq/

wader:
@thushan thanks, good to hear! do you build and try master builds at times? wonder if maybe -rc builds would be interesting at some point?

thushan:
Ah the current team doesn't but not a bad idea! I did do builds for 0.0.5.

wader:
@thushan Ok, i'm mostly worried about breaking windows as i'm not using it myself. Maybe first step could be to try get the expect-based CLI tests running on windows CI also.

thushan:
@wader let me help you with that this week.

wader:
That would be great. If you want to look into running expect here is the CI env it runs in https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md

The expect test is done from here https://github.com/wader/fq/blob/master/Makefile#L24 using https://github.com/wader/fq/blob/master/pkg/cli/test.sh and https://github.com/wader/fq/blob/master/pkg/cli/test.exp. The shell script is helper around the expect script to be able to pass a fq binary to test and also to be silent on success and print log if not.

The windows image has make and bash so hopefully it is mostly about installing expect and then checking if expect.exe is in path?

wader:
Maybe should starta new issue about it to not disturb people

dundee:
0.0.6 just uploaded to Debian unstable.

wader:
@dundee 👍 i saw that fq got added debian testing. Will the version in unstable end up in testing after a while until stable is released or how does that work?

dundee:
Yes, it will migrate to testing in couple of days and then be released in stable someday.

wader:
Great, thanks. Interesting to see how much debian is a gateway into other distros :) if you haven't seen repology before it's a nice tool to keep track of things https://repology.org/project/fq-inspect-binary-data/versions

dundee:
Yes, Debian is used by many other distros as a base. 

It's also good to consider packaging for brew and snap.

wader:
Yeap i have my own brew formula at least and i've looked into having an official one but got a bit stuck on something... also there seems to be an "fq" formula already so have to figure what they do about name conflicts.

For snap there an issue #98 with some ideas but has stalled a bit.

PR #190, opened 2022-03-09
doc: Add bts2022 video

wader:


wader:
@Akaame @xentripetal video of the fq talk at bts2022

PR #240, opened 2022-04-26
Parse all blocks in blkdat files

mutatrum:
Not sure if I have the right construct. When trying this out, it hard locked my machine for a good 5 minutes.

### DIFF
--- a/format/all/all.go
+++ b/format/all/all.go
@@ -9,6 +9,7 @@ import (
 	_ "github.com/wader/fq/format/av1"
 	_ "github.com/wader/fq/format/avro"
 	_ "github.com/wader/fq/format/bencode"
+	_ "github.com/wader/fq/format/blkdat"
 	_ "github.com/wader/fq/format/bson"
 	_ "github.com/wader/fq/format/bzip2"
 	_ "github.com/wader/fq/format/cbor"
--- a/format/blkdat/blkdat.go
+++ b/format/blkdat/blkdat.go
@@ -43,81 +43,83 @@ func decodeBlkDat(d *decode.D, in interface{}) interface{} {
 		for !d.End() {
 			d.FieldStruct("block", func(d *decode.D) {
 	
-			d.FieldU32("magic", scalar.UToSymStr{
-				0xf9beb4d9: "mainnet",
-				0x0b110907: "testnet3",
-				0xfabfb5da: "regtest",
-			}, scalar.Hex, d.AssertU(0xf9beb4d9, 0x0b110907, 0xfabfb5da))
+				d.FieldU32("magic", scalar.UToSymStr{
+					0xf9beb4d9: "mainnet",
+					0x0b110907: "testnet3",
+					0xfabfb5da: "regtest",
+				}, scalar.Hex, d.AssertU(0xf9beb4d9, 0x0b110907, 0xfabfb5da))
 
-			d.Endian = decode.LittleEndian
+				d.Endian = decode.LittleEndian
 
-			size := d.FieldU32("size")
+				size := d.FieldU32("size")
 
-			d.FramedFn(int64(size)*8, func(d *decode.D) {
-				d.FieldStruct("block_header", func(d *decode.D) {
-					d.FieldU32("version")
-					d.FieldRawLen("previous_block_hash", 32*8)
-					d.FieldRawLen("merkle_root", 32*8)
-					d.FieldU32("time")
-					d.FieldU32("bits")
-					d.FieldU32("nonce")
+				d.FramedFn(int64(size)*8, func(d *decode.D) {
+					d.FieldStruct("block_header", func(d *decode.D) {
+						d.FieldU32("version")
+						d.FieldRawLen("previous_block_hash", 32*8)
+						d.FieldRawLen("merkle_root", 32*8)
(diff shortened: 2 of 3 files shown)


wader:
Hey, can you post or link to your test file? The one i tested with is https://github.com/ragestack/blockchain-parser/blob/master/blocks/blk_example.dat and it decodes 2735 transactions in the first block.
Didn't notice that one blkdat file can have several blocks, thanks for noting that. After adding that the test file seems to decode fine with no unknown bits left 🥳 pushed a updated version to #239. But there are probably lots of more things that can be improved

Sorry it locked up your machine :) i guess what happened is that fq ends up using a lot of memory for each of the tx_count fields. I usually do do the ctrl-z/kill %1 when it happens.

Some example queries:
```sh
# last transcation as fancy hex and json
$ go run fq.go -o line_bytes=10 '.[-1].transactions[-1] | ., tovalue' blk_example.dat
        │00 01 02 03 04 05 06 07 08 09│0123456789│.[11].transactions[1462]{}: transaction
0xc25a42│         02 00 00 00         │   ....   │  version: 2
0xc25a42│                     01      │       .  │  input_count: 1
0xc25a42│                        a9 a7│        ..│  inputs[0:1]:
0xc25a4c│8b 31 d7 db 4d ff 73 0a 1d 14│.1..M.s...│
0xc25a56│a2 a7 29 16 a9 61 52 49 fb 4d│..)..aRI.M│
*       │until 0xc25adc.7 (147)       │          │
0xc25ad8│               02            │     .    │  output_count: 2
0xc25ad8│                  cc c8 0b 00│      ....│  outputs[0:2]:
0xc25ae2│00 00 00 00 17 a9 14 e1 0b 02│..........│
0xc25aec│a7 80 2a cc eb 06 d6 86 92 d9│..*.......│
*       │until 0xc25b1f.7 (66)        │          │
0xc25b1e│      99 44 09 00│           │  .D..│   │  locktime: 607385
{
  "input_count": 1,
  "inputs": [
    {
      "scriptsig": "<106>RzBEAiAhu87C9OfAsW3pbmptcaLRDullN8EMGePHamlz2XTdIgIgDwLHWAzXrf4Q+icKofhKQY36IkzimU9FB4jU17xo30IBIQKTJI8bSQboYcEsvCEk7zIkE+Pzh0tb+EGjn7oXM31mBw==",
      "scriptsig_size": 106,
      "sequence": 4294967293,
      "txid": "<32>qaeLMdfbTf9zCh0UoqcpFqlhUkn7TV4pJL6egrfo1zg=",
      "vout": 1
    }
  ],
  "locktime": 607385,
  "output_count": 2,
  "outputs": [
    {
      "scriptpub": "<23>qRThCwKngCrM6wbWhpLZsWnZTsCTE4c=",
      "scriptpub_size": 23,
      "value": "<8>zMgLAAAAAAA="
    },
    {
      "scriptpub": "<25>dqkUUbKQZmrKttYROy/en+FjeyaKIm+IrA==",
      "scriptpub_size": 25,
      "value": "<8>I4hCAQAAAAA="
    }
  ],
  "version": 2
}

# total number of transcations
$ go run fq.go 'map(.transactions | length) | add' blk_example.dat
24068
```

mutatrum:
blk files are capped at ~128MB, currently there are a bit over 3000 of those files, totalling 376GB. I just picked the first file in my local Bitcoin Core blocks folder.

Basically it's an out-of-order dump of the downloaded blocks. If you run Bitcoin Core, it'll download them for you.

I don't know if fq can flush operations, to minimize memory load. Each block is a separate binary blob, that just happens to be in the same file.

mutatrum:
I tried it on blk00000.dat, but there's something not aligned yet.
image

wader:
> blk files are capped at ~128MB, currently there are a bit over 3000 of those files, totalling 376GB. I just picked the first file in my local Bitcoin Core blocks folder.

Aha, that makes sense. 

> Basically it's an out-of-order dump of the downloaded blocks. If you run Bitcoin Core, it'll download them for you.

Ok!

> I don't know if fq can flush operations, to minimize memory load. Each block is a separate binary blob, that just happens to be in the same file.

Nope it's quite simple for now, it decodes using DSL:ish go code that can seek and jump around and it saves metadata about bit range and decoded value for scalars. It would probably be possible to do lazy/deferred decoding for some parts if either you can know the size or you do kind of a first-pass with the decoder and somehow smartly save some ranges. Another thing that could save memory would be to somehow save decoded values in smarter ways.

One issue with lazy/deferred decoding would be that it probably would make queries slower as it have to decode and I/O while evaluating.

wader:
> image

Nice! i haven't focused that much on speed for fq yet but maybe i'll take the profiler for spin on these files to see what is going on.

wader:
> Each block is a separate binary blob

Forgot to say, yes we could split the blkdat format into two formats, blkddat and blkddat_block etc, and then blkdat decoder would just iterate and decode using a "sub" blkdat_block decoder. One problem is what to do if someone wants the forth block in a file, `fq '.[3]' file`, you would still have to decode most parts of the blocks before it to know the offset it is at. btw if you know the offset already you could do, `fq -d raw 'tobytes[12345:] | blkdat_block' file`.

wader:
Let's continue talking in #239

Issue #23, opened 2021-12-23
Add UEFI support


xaionaro:
Just in case wondering: does it make sense to add support of UEFI files (using https://github.com/linuxboot/fiano)?

Asking just because it might be a nice onboarding task for junior engineers.

wader:
Hi, i don't know much about UEFI. Is it similar to an executable format like ELF? haven't look close at the go package, for fq to be able to use it you probably have to have some kind of API that supports iterate thru strucutres or some way of accessing offsets/lengths where things are located.

xaionaro:
> Is it similar to an executable format like ELF?

It is a tree-like structure, where roughly speaking on the leafs it may contain PE32 or other files.

![Screenshot from 2021-12-23 15-11-19](https://user-images.githubusercontent.com/2349376/147259089-93ebc8d4-6b35-4976-ada5-f5d001f7eb32.png)

>  for fq to be able to use it you probably have to have some kind of API that supports iterate thru strucutres or some way of accessing offsets/lengths where things are located.

Yeah, this is exactly what [fiano](https://github.com/linuxboot/fiano) package is about for UEFI layouts.

wader:
Ok! Interesting, and PE32 here is a 32-bit portable executable (windows exe)?

Don't have much to look into it right now but it would be helpful for me or someone who wants to work on it to have some resources like small examples files or links to tools to create own files.

xaionaro:
> Interesting, and PE32 here is a 32-bit portable executable (windows exe)?

Yeaaaaah. I also was a bit shocked when found out they use EXE files inside UEFI images. Though it works different there, but the executable container is the same, yes.

> Don't have much to look into it right now but it would be helpful for me or someone who wants to work on it to have some resources like small examples files or links to tools to create own files

At this stage I was just wondering if this makes sense, just to keep this opportunity in mind :)

wader:
Feels like it could be a good fit. Also fq has infrastructure to do "subformat" decoding so once where is a PE decode (or other formats in UEFI) it could automatically do nested decoding. Im currently working on a much improved ELF decoder (https://github.com/wader/fq/tree/elf if you want to sneak peak) and i hope PE is not that much different, but i've heard it's quite legacy heavy :)

tlehman:
I am working on adding MBR support for looking at the 512 byte section of bootable iso images. I'll look into UEFI next since that's the successor to the legacy MBR BIOS booting scheme.

wader:
[...]

PR #241, opened 2022-04-27
Add MBR format

tlehman:
Related to issue #23 

This is a first pass, it breaks down the MBR into the code_area, the partition_table, and the signature (magic number denoting the end of the boot record)

## Future work
- Disassemble the 16-bit opcodes in the code area
for example
```sh
objdump -D -Mintel,i8086 -b binary -m i386 format/mbr/testdata/mbr.bin | head -15

format/mbr/testdata/mbr.bin:     file format binary


Disassembly of section .data:

00000000 <.data>:
   0:	fa                   	cli
   1:	bc 00 7c             	mov    sp,0x7c00
   4:	31 c0                	xor    ax,ax
   6:	8e d0                	mov    ss,ax
   8:	8e c0                	mov    es,ax
   a:	8e d8                	mov    ds,ax
   c:	52                   	push   dx
   d:	be 00 7c             	mov    si,0x7c00
```

# Background
MBR is short for Master Boot Record. It is the legacy method of booting,
the code area is limited to 446 bytes, and it's all 16-bit x86 opcodes.
Old BIOS code knows how to call into it and execute the instructions there.
MBR also has a 64-byte partition table that can store up to 4 partitions.
The GPT partitioning scheme obsoletes this, and allows for more partitions, but
that requires UEFI to work.

For more information on MBR, see: https://thestarman.pcministry.com/asm/mbr/PartTables.htm#mbr


format/mbr/mbr.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/wader/fq/format"
+	"github.com/wader/fq/format/registry"
+	"github.com/wader/fq/pkg/decode"
+)
+
+func init() {
+	registry.MustRegister(decode.Format{
+		Name:        format.MBR,
```
wader, line 11: Are there other computer platforms that use the name "mbr"? thinking maybe pc_mbr etc? but maybe mbr is synonomus with IBM PC MBR?

tlehman: Good question, I'll look if there are any collisions and if so, disambiguate

doc/dev.md
```diff
@@ -5,7 +5,7 @@
 - Create a directory  `format/<name>`
 - Copy some similar decoder, `format/format/bson.go` is quite small, to `format/<name>/<name>.go`
 - Cleanup and fill in the register struct, rename `format.BSON` and add it
-to `format/fromat.go` and don't forget to change the string constant.
+to `format/format.go` and don't forget to change the string constant.
```
wader, line 8: 👍

format/mbr/testdata/mbr.fqtest
```diff
@@ -0,0 +1,9 @@
+# head -c 512 /dev/sda > mbr.bin
+$ fq . -d mbr mbr.bin
+     |00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f|0123456789abcdef|.{}: mbr.bin (mbr)
+0x000|fa bc 00 7c 31 c0 8e d0 8e c0 8e d8 52 be 00 7c|...|1.......R..||  code_area: raw bits
+*    |until 0x1bd.7 (446)                            |                |
+0x1b0|                                          00 00|              ..|  partition_table: raw bits
+0x1c0|00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00|................|
+*    |until 0x1fd.7 (64)                             |                |
+0x1f0|                                          55 aa|              U.|  boot_record_sig: raw bits
```
wader, line 9: This is always 0x55aa for a mbr? if so could seek and assert i guess

format/mbr/mbr.go
```diff
@@ -0,0 +1,22 @@
...
+		Description: "Master Boot Record",
+		DecodeFn:    mbrDecode,
+	})
+}
+
+func mbrDecode(d *decode.D, in interface{}) interface{} {
+	d.FieldRawLen("code_area", 446*8)
```
wader, line 18: Plan is to use some kind of x86_16 decoder here?

format/mbr/mbr.go
```diff
@@ -0,0 +1,22 @@
...
+	d.FieldRawLen("partition_table", 64*8)
```
wader, line 19: Know if this partition table format is used outside mbr? if not i guess maybe not worth having it as separate format

wader:
Nice start! will be interesting to get started with ISA decoding in fq, have some ideas but nothing clear yet so we will probably have to try things out :)

wader:
Just comment here, IM or send me an email if you have any questions or want to discuss something

wader:
Tried the x86_16 decoder, seems to work ok. But i'm not that pleased with the current output, it would be nice to split out more operands and such.. hmm

```sh
$ go run fq.go -o line_bytes=4 -d raw 'tobytes[0:446] | x86_16 | dd' mbr.bin
     │00 01 02 03│0123│.[0:193]: (x86_16)
     │           │    │  [0]{}:
0x000│fa         │.   │    opcode: "cli" (raw bits)
     │           │    │    op: "cli" (0xfa000000)
     │           │    │  [1]{}:
0x000│   bc 00 7c│ ..|│    opcode: "mov sp, 0x7c00" (raw bits)
     │           │    │    op: "mov" (0xbc000000)
     │           │    │  [2]{}:
0x004│31 c0      │1.  │    opcode: "xor ax, ax" (raw bits)
     │           │    │    op: "xor" (0x31c00000)
     │           │    │  [3]{}:
0x004│      8e d0│  ..│    opcode: "mov ss, ax" (raw bits)
     │           │    │    op: "mov" (0x8ed00000)
     │           │    │  [4]{}:
0x008│8e c0      │..  │    opcode: "mov es, ax" (raw bits)
     │           │    │    op: "mov" (0x8ec00000)
     │           │    │  [5]{}:
0x008│      8e d8│  ..│    opcode: "mov ds, ax" (raw bits)
     │           │    │    op: "mov" (0x8ed80000)
     │           │    │  [6]{}:
0x00c│52         │R   │    opcode: "push dx" (raw bits)
     │           │    │    op: "push" (0x52000000)
     │           │    │  [7]{}:
0x00c│   be 00 7c│ ..|│    opcode: "mov si, 0x7c00" (raw bits)
     │           │    │    op: "mov" (0xbe000000)
     │           │    │  [8]{}:
0x010│bf 00 06   │... │    opcode: "mov di, 0x600" (raw bits)
     │           │    │    op: "mov" (0xbf000000)
....
```

format/mbr/testdata/mbr.fqtest
tlehman: Yes, I'll do that

tlehman:
Probably going to pick up work on this next week, it's Hack Week at SUSE

wader:
🥳 let me know if you have questions or want to discuss something

format/mbr/mbr.go
wader: 👍

format/mbr/testdata/mbr.fqtest
```diff
@@ -0,0 +1,9 @@
+# head -c 512 /dev/sda > mbr.bin
+$ fq . -d mbr mbr.bin
```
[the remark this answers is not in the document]
tlehman: I'm getting this error when I type `go run fq.go '.boot_record_sig' -d mbr dv format/mbr/testdata/mbr.bin`

```
error: dv: no such file or directory
     |00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d|0123456789abcd|
0x1f8|                  55 aa|                 |      U.|     |.boot_record_sig: raw bits
exit status 2
```

I had it in my notes from April 26 that the dv was passed in after the <format> arg, I think I wrote that down wrong. What's the correct syntax for getting verbose output?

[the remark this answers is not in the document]
wader: Yes it's `dv`. I think the issue is that "dv" is seen as the first input filename (format/mbr/testdata/mbr.bin the second input). Try `go run fq.go '.boot_record_sig | dv' -d mbr format/mbr/testdata/mbr.bin`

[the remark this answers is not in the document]
wader: fq's (and jq) arguments work by having one expression and evaluate it on multiple inputs (unless you do -n)

format/mbr/mbr.go
tlehman: For the first iteration I'm going to focus on the partition table. Your work on the x86_16 decoder can be reused here.

wader: Sounds good 👍

format/mbr/mbr.go
```diff
@@ -0,0 +1,66 @@
...
+	d.FieldStruct("entry_4", decodePartitionTableEntry)
+}
+
+// Source: https://thestarman.pcministry.com/asm/mbr/PartTables.htm#Decoding
+func decodeCHSBytes(d *decode.D) scalar.S {
+	head, _ := d.Bits(8)
+	sectorHighBits, err := d.Bits(2)
+	if err != nil {
+		d.IOPanic(err, "chs")
+	}
```
wader, line 48: You can simplify this to just `d.U2()` or `d.U(2)` and it will do the error check and panic. I should probably rename´Bits` to `TryBits`, the convention usually is `Try*` functions returns error other just panic and there is usually pairs of functions.

format/mbr/mbr.go
```diff
@@ -0,0 +1,66 @@
...
+func decodePartitionTableEntry(d *decode.D) {
+	d.FieldU8("boot_indicator", scalar.UToDescription{
+		0x80: "active",
+		0x00: "inactive",
+	})
+	d.FieldStrScalarFn("starting_chs_vals", decodeCHSBytes)
+	d.FieldU8("partition_type", partitionTypes)
+	d.FieldStrScalarFn("ending_chs_vals", decodeCHSBytes)
+	d.FieldStrScalarFn("starting_sector", decodeCHSBytes)
+	d.U8() // extra byte
```
wader, line 29: Maybe field for this byte also? otherwise it will just be skipped and end up as "unknown"

format/mbr/mbr.go
```diff
@@ -0,0 +1,66 @@
...
+		Name:        format.MBR,
+		Description: "Master Boot Record",
+		DecodeFn:    mbrDecode,
+	})
+}
+
+func decodePartitionTableEntry(d *decode.D) {
+	d.FieldU8("boot_indicator", scalar.UToDescription{
+		0x80: "active",
+		0x00: "inactive",
```
wader, line 23: You can do a bit field struct here (struct with `d.FieldBool("...")`) if it's useful to be able to address individual bits. Sometimes i also use `0b1000_0000` etc literals for binary literals

format/mbr/mbr.go
```diff
@@ -0,0 +1,66 @@
...
+		d.IOPanic(err, "chs")
+	}
+	sector, _ := d.Bits(6)
+	cylinderLowerBits, err := d.Bits(8)
+	if err != nil {
+		d.IOPanic(err, "chs")
+	}
+	cylinder := (sectorHighBits << 2) | cylinderLowerBits
+	return scalar.S{Actual: fmt.Sprintf("CHS(%x, %x, %x)", cylinder, head, sector)}
+}
```
wader, line 56: Could it make sense to read CHS as a struct with cylinder, head , sector fields?

doc/formats.md
```diff
@@ -352,6 +353,16 @@ Return `matroska_path` string for a box decode value
...
+### mbr
+
+The MBR (Master Boot Record) tells the computer how to find the bootloader.
+It is a fixed 512-byte record that PC motherboards know how to read.
+It contains 446 bytes of executable 16-bit x86 code, and a 64-bit partition
+table, then two magic bytes to signify the end of the record.
+
+#### References and links
+- https://thestarman.pcministry.com/asm/mbr/PartTables.htm#mbr
+
```
wader, line 365: This has changed a bit in master since the PR, was redone so that documentation can embedded into fq, accessible via `help(...)` and `--help`. Ex how it works now:
https://github.com/wader/fq/blob/master/format/asn1/asn1_ber.jq#L10
https://github.com/wader/fq/blob/master/format/asn1/asn1_ber.go#L40

And then `make doc` will generate things for `formats.md` also

wader:
Cleaned up some of the Try* things https://github.com/wader/fq/pull/302 if you switch to `d.*U*()` instead of `d.Bits` it should be fine.

wader:
Also fixed some of the doc typos https://github.com/wader/fq/pull/303

format/mbr/mbr.go
tlehman: I haven't found any other formats that use the "MBR" as their initialism. In fact, I learned that non-x86 machines like RPi use the same exact format but just ignore the code section and only use the partition table.

doc/formats.md
tlehman: I added a `format/mbr/mbr.jq` file with the `_mbr__help` function, and then added `Functions: []string{"_help"}` to the registry.MustRegister call in mbr.go. Before doing this, `make` built the fq binary after all the tests passed. With the _help stuff, it fails with 

```
expr:1:0: compile: function not defined: _mbr__help/0
```

tlehman: yep, that was it

format/mbr/mbr.go
tlehman: Yes, and that suggestion is in line with [this general tip](https://github.com/wader/fq/blob/master/doc/dev.md#some-general-tips:~:text=Try%20to%20not%20decode%20too%20much%20as%20one%20value.%20A%20length%20encoded%20int%20could%20be%20two%20fields%2C%20but%20maybe%20a%20length%20prefixed%20string%20should%20be%20one.%20Flags%20can%20be%20struct%20with%20bit%2Dfields.)

format/mbr/mbr.go
wader: Aha didn't know, so ignore code section by leaving it zero etc?

Yeap let's call it mbr 👍

format/mbr/mbr.go
wader: Maybe this could use `scalar.UToSym` that way you can write queries as `.boot_indicator  == "active"` and if you really want to compare the non-mapped value do `(.boot_indicator | toactual) == 0x80`

format/mbr/mbr.go
wader: Ah yes, forgot i wrote that :) maybe should add something more about that it also makes it easier to write queries if things are split up into more individual fields

format/mbr/mbr.go
```diff
@@ -0,0 +1,72 @@
...
+	sectorHighBits, err := d.Bits(2)
+	if err != nil {
+		d.IOPanic(err, "chs")
+	}
+	sector, _ := d.Bits(6)
+	cylinderLowerBits, err := d.Bits(8)
+	if err != nil {
+		d.IOPanic(err, "chs")
+	}
+	cylinder := (sectorHighBits << 2) | cylinderLowerBits
```
wader, line 60: Oh the cylinder bits are "continuous" as 16LE, this is a bit messy to represent :( i've run into this issue with some other formats like https://github.com/wader/fq/blob/master/format/vpx/vp8_frame.go#L36 haven't figured any nice way to handle this yet. Also would be nice if there was some decoding helpers for doing this

format/mbr/mbr.jq
```diff
@@ -0,0 +1,7 @@
+def _mbr__help:
+    { notes: "Supports decoding Master Boot Record data",
```
wader, line 2: Looking at the generated help text maybe this is redundant? I've used this to note missing/extra features etc

format/mbr/mbr.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/wader/fq/format/registry"
+	"github.com/wader/fq/pkg/decode"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+//go:embed mbr.jq
+var mbrFS embed.FS
+
+func init() {
+	registry.MustRegister(decode.Format{
```
wader, line 17: This was refactored a bit in master, is now `interp.RegisterFormat(...)`

tlehman: Will rebase and fix

wader:
Just a note if you start working on this again: in master format documentation is now .md files again, so you can convert format/mbr/mbr.jq into a normal markdown file, see macho.md etc, and then embed it, see macho.go etc. Then it will be used both for generating documentation and cli help.

wader:
Format registration on master have changed a bit but should be straight forward to copy/paste from msgpack.go etc and rename some things and also do the same for formats.go

PR #301, opened 2022-06-26
Update usage.md

ksa-real:


doc/usage.md
```diff
@@ -280,7 +280,7 @@ This type is returned by decoders and it used to represent parts of the decoed i
 
 Each decode value has these properties:
 - A bit range in the input
-  - Can be accessed as a binary using `tobits`/`tobytes`. Use the `start` and `size` keys to postion and size.
+  - Can be accessed as a binary using `tobits`/`tobytes`. Use the `start` and `size` keys to position and size.
```
wader, line 283: Thanks, maybe should even say ”to access” now when i read it?

ksa-real: Sure, somehow this part hasn't triggered me :)

wader: Same for me at times :) so thanks for reading and correcting. Should really spend time more on the documenatation for fq, usually very useful to try explain things

wader:
Thanks, i will fix the other thing i mentioned later on.

PR #336, opened 2022-07-23
avro_ocf: Fix panic on missing meta schema

wader:


wader:
@xentripetal is ok? had a quick look around for other type asserts but looked ok i think

Ran into this while stress testing fq a bit

wader:
I'll merge, let me know if it should be changed somehow

PR #352, opened 2022-08-02
macho: Split into macho/macho_fat, fix offset issue and add string de…

wader:
…coding

Split fat macho into own decoder macho_fat. This also fixes issue with section
offset etc not being correct as they are from the start of each embedded file.

Make all address and offset field be in hex.

Decode __cstring, __ustring and __cfstring sections.

Fix LC_ENCRYPTION_INFO_64 missing pading issue.

Skip ranging for __bss and __common as they dont have any data in the file.

Simplifed magic handling a bit and add symbols.

Simplified state struct field, had redudant struct.


wader:
@Akaame noticed fat macho acted a bit weird with offsets (offset is in each file not start of fat file), ended up refactor a bit and also added string section decoding.

PR #378, opened 2022-08-17
Add sample size entry to list for 32bps flac streams

ktmf01:
It seems nothing else is necessary to implement decoding of 32bps FLAC.

See also https://github.com/wader/flac.tcl/pull/1


wader:
Thanks!

If you have a ffmpeg that is 32bit capable you can maybe modify https://github.com/wader/fq/blob/master/format/flac/testdata/gen to generate test cases? but we can merge and do it later also.

ktmf01:
I've created the FLAC files, but I'm not sure how to generate proper fqtest files? When simply running `sh gen` in the directory format/flac/testdata I get fqtest files with only a fq command in them, not an fq output. Running the whole test suite here fails

```
~/bin/fq $ make test fq
go test -race   ./format ./internal/aheadreadseeker ./internal/ansi ./internal/asciiwriter ./internal/columnwriter ./internal/ctxreadseeker ./internal/hexdump ./internal/hexpairwriter ./internal/mathextra ./internal/recoverfn ./internal/shquote ./pkg/bitio ./pkg/interp ./pkg/ranges
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/format	0.011s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/aheadreadseeker	0.018s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/ansi	0.021s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/asciiwriter	0.009s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/columnwriter	0.010s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/ctxreadseeker	0.012s
?   	github.com/wader/fq/internal/hexdump	[no test files]
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/hexpairwriter	0.010s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/mathextra	0.013s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/recoverfn	0.011s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/internal/shquote	0.011s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/pkg/bitio	0.008s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/pkg/interp	0.008s
FATAL: ThreadSanitizer: unsupported VMA range
FATAL: Found 39 - Supported 48
FAIL	github.com/wader/fq/pkg/ranges	0.008s
FAIL
make: *** [Makefile:18: testgo] Fout 1
```

wader:
> I've created the FLAC files, but I'm not sure how to generate proper fqtest files? When simply running `sh gen` in the directory format/flac/testdata I get fqtest files with only a fq command in them, not an fq output. Running the whole test suite here fails

Ah sorry i should have explained how it works. `.fqtest` files can be used in two mode, verify or write actual output, so the `gen` script will generate flac files and also "partial" .fqtest files that needs to be run in "write" mode first. You can do something like this to run in write mode and only run the flac tests:

```
$ WRITE_ACTUAL=1 go test -v -run TestFQTests/flac ./format/
```

That will modify the `.fqtest` files and add current output. If you run without `WRITE_ACTUAL=1` it will verify the output instead and also provide a diff if it fails.

> ```
> ~/bin/fq $ make test fq
> ...
> FATAL: ThreadSanitizer: unsupported VMA range
> FATAL: Found 39 - Supported 48
> FAIL	github.com/wader/fq/internal/aheadreadseeker	0.018s
> ...
> ```

Hmm think i've seen this happen when running on some env were golang does not support `-race` race condition detector. Is this in a emulated env or some container? but you can probably skip `-race`, if you'r not messing around with internal packages or the interpreter you can run all format tests using `go test ./format/`.

BTW if you get some test failing on string escaping being different `\u00##` vs `\x##` if because %q formatter was changed in go 1.19. I should probably workaround that somehow.

wader:
Think i see what i going on now, the gen script it very old, things have changed. Can you rebase on master once https://github.com/wader/fq/pull/381 is merged?

Hmm strange about the ThreadSanitizer, arm64 should be supported for `-race`, will digg a bit

wader:
Seems to be some  mmu configurations that ThreadSanitizer don't like https://forums.raspberrypi.com/viewtopic.php?t=275372&start=925#p1903632 but i think you can just skip running with `-race` for now and i now also remember that there is a workaround in the Makefile that one of the CI-builds use do skip it `GO_TEST_RACE_FLAGS= make test`

ktmf01:
Change has been rebased. I've changed some things in gen: for example, I've added two resample filters to the ffmpeg filter chain. This is because some of the generators produce 16-bit PCM samples, which do not really test the 32-bit PCM processing as this results in 16 wasted bits.

One more note: the generated 32-bit stereo testfile does not use any stereo decorrelation. So, this test isn't very thorough at the moment.

wader:
Ok great! Let's merge and feel free to add more tests later

PR #379, opened 2022-08-18
Fix decoding of FLAC raw entropy partition

ktmf01:
As was mentioned in the TODO of format/flac/flac.go

Please review. Go is new to me, and I don't understand the syntaxis used here, simple copy-paste-trial-error approach.

I think the problem was that (as in [flac.tcl, samples were read unsigned, but](https://github.com/wader/flac.tcl/pull/2)) samples were read unsigned instead of signed, but I really don't understand what `d.FieldRawLen` does, so maybe it didn't read any samples at all?

### DIFF
--- a/format/flac/flac_frame.go
+++ b/format/flac/flac_frame.go
@@ -463,7 +463,13 @@ func frameDecode(d *decode.D, in interface{}) interface{} {
 
 								if riceParameter == riceEscape {
 									escapeSampleSize := int(d.FieldU5("escape_sample_size"))
-									d.FieldRawLen("samples", int64(count*escapeSampleSize))
+									d.RangeFn(d.Pos(), int64(count*escapeSampleSize), func(d *decode.D) {
+										d.FieldRawLen("samples", int64(count*escapeSampleSize))
+									})
+									for j := 0; j < count; j++ {
+										samples[n] = d.S(escapeSampleSize)
+										n++
+									}
 								} else {
 									samplesStart := d.Pos()
 									for j := 0; j < count; j++ {
--- a/format/flac/flac.go
+++ b/format/flac/flac.go
@@ -2,9 +2,6 @@ package flac
 
 // https://xiph.org/flac/format.html
 // https://wiki.hydrogenaud.io/index.php?title=FLAC_decoder_testbench
-// TODO:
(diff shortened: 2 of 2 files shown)


format/flac/flac_frame.go
```diff
@@ -463,7 +463,13 @@ func frameDecode(d *decode.D, in interface{}) interface{} {
...
 								if riceParameter == riceEscape {
 									escapeSampleSize := int(d.FieldU5("escape_sample_size"))
-									d.FieldRawLen("samples", int64(count*escapeSampleSize))
+									d.RangeFn(d.Pos(), int64(count*escapeSampleSize), func(d *decode.D) {
+										d.FieldRawLen("samples", int64(count*escapeSampleSize))
+									})
+									for j := 0; j < count; j++ {
+										samples[n] = d.S(escapeSampleSize)
+										n++
+									}
```
wader, line 472: Aha that was the reason the failing test files produced different samples, nice catch. Before this change escaped samples were not read as samples at all, the bits for them were just added as a raw bit field. `FieldRawLen` reads raw bits without doing any interpretation of them.

wader: BTW we could add an option to optionally add fields for each sample if you think that is useful? the reason i skipped doing it for now is that it would use lots of memory and make the output not very useful.

Then one could for example decode a specific frame with sample fields enabled, something like `.frames[10] | flac_frames({decode_samples: true})`

wader:
LGTM, will merge once tests pass. Thanks!

format/flac/flac_frame.go
ktmf01: `flac` has a similar `--residual-text` option that I've used a few times when debugging some patches. For example, sometimes it wasn't clear whether a bug was in the encoder or decoder, then looking at these values would help. It could be helpful, but it might be too specific for `fq`?

wader: fq is all about debugging and details so is fine i think :) but maybe it would be a flac_frame decoder option that is disable by default (current behaviour). I think it would be more or less just add a `DecodeSamples bool` to `FlacFrameIn` struct (in format.go) etc and initialize it to `false` and then decode conditionally on it. The interpreter code takes care of all the translation of jq function opts and such.

PR #383, opened 2022-08-18
Add WASM format

bitbears-dev:
Hello, I have created this because I personally needed it.
Tested against bunch of .wasm / .fqtest pairs in `format/wasm/testdata/core`. (The .wasm files are generated from the standard test suite i.e. github.com/WebAssembly/spec/test/core.)

I would be happy to receive any feedback.

Thank you!


wader:
Hey, nice! Yes will review as soon i get time, later today most probably.

Surely looks like there is not a lack of tests 😄 

What was you personal need for if that is ok to ask? might help review, usage help and ideas for how to decode things

wader:
Sadly it seems the lint action does not add annotation for some reason, will have to look into that, but have a look at the action log what is going on. Alternatively you can run `make lint` locally also.

format/wasm/instructions.go
```diff
@@ -0,0 +1,1539 @@
...
+		d.FieldU8(name, scalar.Sym("ε"))
+	case 0x6f, 0x70, 0x7b, 0x7c, 0x7d, 0x7e, 0x7f:
+		d.FieldU8(name, valtypeToSymMapper)
+	default:
+		d.FieldSScalarFn(name, readSignedLEB128)
+	}
+}
+
+func decodeMemArg(d *decode.D, name string) {
+	d.FieldStruct(name, func(d *decode.D) {
```
wader, line 21: Can add `//nolint:nolint` if it's ok

format/wasm/mappers.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+var valtypeToSymMapper = &valtypeToSym{}
+
+type valtypeToSym struct {
+}
+
+// valtype ::= t:numtype => t
+//          |  t:vectype => t
+//          |  t:reftype => t
```
wader, line 61: Maybe a go 1.19 godoc fmt thing?

format/wasm/mappers.go
```diff
@@ -0,0 +1,211 @@
...
+	case 0x7c:
+		s.Sym = "f64"
+	case 0x7b:
+		s.Sym = "v128"
+	case 0x70:
+		s.Sym = "funcref"
+	case 0x6f:
+		s.Sym = "externref"
+	default:
+		s.Sym = "unknown valtype"
```
wader, line 84: Some of these mappers might be possible to express using the predefined table mappers
```go
var valtypeMapper = scalar.UToSymStr{
	0x7f:   "i32",
        // ...
}
```

wader: Maybe could be nice to add constant for the number literals? but it depends a bit i guess, i usually add constant if the number(s) is used in severals places or if i want to do something more fancy with them

format/wasm/testdata/core/generate_fqtest.sh
```diff
@@ -0,0 +1,48 @@
...
+
+  go build -o "$tmpdir/fq" "$d/../../../../"
+  find "." -name '*.wasm' -print0 | sort -z -n |
+    while IFS= read -r -d '' f; do
+      f=${f#./}
+      of=${f/%.wasm/.fqtest}
+      echo "$f -> $of"
+      echo "$ fq -d wasm dv $f" > "$of"
+      "$tmpdir/fq" -d wasm dv "$f" >>"$of"
+    done
```
wader, line 45: You can run the fqtest tests in "write" mode and they will write instead of verify the output, ex something like this would rewrite all .fqtest files with current actual output `WRITE_ACTUAL=1 go test -run TestFQTests/wasm ./format/`. So what i usually do it so just generate .fqtest files with `$ fq ...`  lines etc and then run in write mode.

format/wasm/mappers.go
```diff
@@ -0,0 +1,211 @@
...
+
+	switch v {
+	case sectionIDCustom:
+		s.Sym = "custom section"
+	case sectionIDType:
+		s.Sym = "type section"
+	case sectionIDImport:
+		s.Sym = "import section"
+	case sectionIDFunction:
+		s.Sym = "function section"
```
wader, line 28: I've usually tried to use string symbol values that look "codey", so no whitespace and maybe lowercase. But it might depend a bit on what the spec uses the names etc?

format/wasm/instructions.go
```diff
@@ -0,0 +1,1539 @@
...
+
+func decodeSelectT(d *decode.D) {
+	d.FieldU8("select", d.AssertU(0x1c), scalar.ActualHex)
+	decodeVec(d, "t", func(d *decode.D) {
+		decodeValType(d, "t")
+	})
+}
+
+func decodeLocalGet(d *decode.D) {
+	d.FieldU8("local.get", d.AssertU(0x20), scalar.ActualHex)
```
wader, line 548: Probably a good idea to only use valid jq identifiers as field name, ex: `.something.local_get` vs `.something."local.get"` or `.something.["local.get"]`

format/wasm/instructions.go
```diff
@@ -0,0 +1,1539 @@
...
+
+func decodeVectorInstruction(d *decode.D) {
+	d.FieldU8("prefix", d.AssertU(0xfd), scalar.ActualHex)
+	s := peekUnsignedLEB128(d)
+	v, ok := s.Actual.(uint64)
+	if !ok {
+		d.Fatalf("expected uint64 but got %t", s.Actual)
+	}
+	switch v {
+	case 0:
```
wader, line 854: For some formats it has sometimes been helpful to have custom type with names and a decode function and then create a table with decoders. But i like simple switch statements also sometimes.

Some inspiration is https://github.com/wader/fq/blob/master/format/msgpack/msgpack.go and maybe https://github.com/wader/fq/blob/master/format/cbor/cbor.go

wader:
GitHub does not allow me to review `wasm.go`, too many files it seems :) I noticed it had a couple of LEB128 functions, wonder if they could be refactored out and shared, there are some other formats using LEB128 also i think. But we can merge and i can do various refactorings later.

wader:
Did an initial review so that we can get started, will look more later

Great stuff 👍

bitbears-dev:
Thank you for your prompt review! All of the points you raised are valid and will be addressed immediately.

format/wasm/instructions.go
bitbears-dev: added `//nolint:unparam` @ https://github.com/wader/fq/pull/383/commits/3fca7cc09a434fd4ba657be69c39934d7be826d0

format/wasm/mappers.go
bitbears-dev: modified to use `scalar.UToSymStr` @ https://github.com/wader/fq/pull/383/commits/63f4a7267c3f4920bea98baea736663831afa417

It's way simple!

format/wasm/mappers.go
bitbears-dev: Section IDs are defined in the specification like:
image

But I agree that there is benefit in having the symbol values be ”codey". I'll change them to use underscores, not white spaces.

format/wasm/instructions.go
bitbears-dev: I'll change the field name to `opcode` or something, and have the mnemonic as symbol value.

format/wasm/instructions.go
bitbears-dev: I like the idea to have such a custom type. I'll consider it.

bitbears-dev:
> What was you personal need for if that is ok to ask? might help review, usage help and ideas for how to decode things

I'm writing a small ebook to explain the data format of WebAssembly binary modules. I would like to use the output from `fq` in the book, and also want readers to try fq's interactive mode to dig into deeper.

bitbears-dev:
> I noticed it had a couple of LEB128 functions, wonder if they could be refactored out and shared, there are some other formats using LEB128 also i think. 

Yeah, I thought the same. AFAIK, LEB128 is from DWARF, so I thought decoder for ELF should have something for it. Then I grepped `leb`, `128` or some other keywords, but found nothing. So I created my own implementation. While I was writing code, I thought "This function may be used some other decoders. Should I place it in some other package like `internal` or `pkg`? But where? :thinking: "  

> But we can merge and i can do various refactorings later.

I agree. Please refactor out the code wherever appropriate.

format/wasm/mappers.go
wader: Yeah they are nice, there are from/to table mappers for most types supported

format/wasm/mappers.go
wader: 👍  if you want a longer text not ment for queries etc then scalars values can have a description also

format/wasm/instructions.go
wader: When writing decoders i usually try to think how to make the structure easy to work with for doing queries etc. For example maybe group things into structs that might be useful as one unit and also try to avoid repetition that would make queries long to write.

format/wasm/instructions.go
wader: Yeap do so but plain non-fancy-pantsy code is nice sometimes too

wader:
> I'm writing a small ebook to explain the data format of WebAssembly binary modules. I would like to use the output from `fq` in the book, and also want readers to try fq's interactive mode to dig into deeper.

Wow nice! :) that sounds great, look forward reading it. I've played around with various idea for documentation for fq, let me know if you want discuss idea. If you want to get nice output from fq (or other CLI tools) i wrote this thingy https://github.com/wader/ansisvg search for ansisvg in the repo for some examples.

Also formats can add their own native go function or jq functions if there are some functionallity like queries and conversions etc. For example would be interesting to try write a wasm to S-expressions function in jq? or some kind of encode support? ....or maybe even a basic interpreter? 🤔 hehe

wader:
Will probably do some more reviewing later today

wader:
The magic seems unique so i think you can add this format to the probe group, just a matter for add a 		`Groups: []string{format.PROBE}` to the `interp.RegisterFormat(...)` struct.

BTW if you run `make doc` the wasm format should be added to README.md, usage.md and some other places.

wader:
The `peekUnsignedLEB128` one is because you and to read a LEB128 and possibly fail but on success the position should not have changed? I'm wondering if i should generate ´Peek*` methods for all decode types?

wader:
There are some type assert helpers for scalar values, ex:
```go
v, ok := s.Actual.(uint64)
if !ok {
	d.Fatalf("expected uint64 but got %t", s.Actual)
}
```
can be:
```go
v := s.ActualU()
```

wader:
The code in instructions.go looks very neat now 👍

wader:
Played around a bit with the decode, feels very stable, fun to learn a bit about wasm.

```sh
$ go run . -d wasm '.sections | {import: map(select(.id == "import_section").content.im.x[].nm.b), export: map(select(.id == "export_section").content.ex.x[].nm.b)}' ~/src/gopacket-wasm/main.wasm
{
  "export": [
    "run",
    "resume",
    "getsp",
    "mem"
  ],
  "import": [
    "debug",
    "runtime.wasmExit",
    "runtime.wasmWrite",
    "runtime.nanotime",
    "runtime.walltime",
    "runtime.scheduleTimeoutEvent",
    "runtime.clearTimeoutEvent",
    "runtime.getRandomData",
    "syscall/js.stringVal",
    "syscall/js.valueGet",
    "syscall/js.valueSet",
    "syscall/js.valueIndex",
    "syscall/js.valueSetIndex",
    "syscall/js.valueCall",
    "syscall/js.valueInvoke",
    "syscall/js.valueNew",
    "syscall/js.valueLength",
    "syscall/js.valuePrepareString",
    "syscall/js.valueLoadString",
    "syscall/js.valueInstanceOf"
  ]
}

$ go run . -d wasm -r '.sections[] | select(.id == "code_section") | [.. | .opcode? // empty] | count | sort_by(.[1]) | .[] | "\(.[1]): \(.[0])"' ~/src/gopacket-wasm/main.wasm
1: f64.neg
1: i32.mul
1: i32.store16
1: i64.trunc_f64_s
1: i64.trunc_f64_u
1: memory.grow
1: memory.size
2: i32.gt_u
2: i32.load8_s
2: i32.store
2: i32.xor
2: i64.div_s
3: i64.extend_i32_s
3: i64.rem_s
4: f64.ge
4: i32.ne
5: i32.load8_u
6: drop
8: i64.rem_u
9: f64.sub
10: f64.add
11: i32.and
12: f32.store
13: f64.gt
13: i64.load16_s
14: f64.lt
15: f64.convert_i64_u
20: f64.div
20: i64.load8_s
28: f64.mul
29: f64.ne
33: f32.load
33: f64.eq
34: f32.demote_f64
42: f64.convert_i64_s
55: f64.promote_f32
64: i32.eq
70: br_if
74: i64.div_u
105: f64.const
108: f64.store
119: i64.ge_u
132: f64.load
166: i64.xor
173: i64.le_s
417: i64.gt_s
519: call_indirect
519: i64.ge_s
521: i32.shr_u
546: i64.store16
580: i64.ne
607: i64.load16_u
664: i64.gt_u
679: i64.mul
802: i64.load32_s
1013: i64.lt_s
1084: i64.shr_s
1789: i64.or
1928: i64.store32
1931: i64.sub
1946: i64.le_u
2037: i64.and
2766: i64.shr_u
3243: i64.store8
3275: i64.eq
3457: i64.lt_u
3526: i32.le_u
3528: i32.load
3623: select
3631: br_table
3703: end
3704: loop
3721: i64.load32_u
4488: nop
4755: i64.shl
6268: i64.load8_u
6613: unreachable
7393: local.tee
9493: i32.add
10825: i32.load16_u
14266: i64.eqz
16332: 32.eqz
24318: call
25291: i32.sub
26846: return
28174: i64.add
30527: br
34634: local.set
41158: i64.extend_i32_u
44977: if
50623: i64.load
54830: block
58266: local.get
68563: i32.wrap_i64
71360: i64.store
99984: global.set
105376: i64.const
116719: i32.const
172257: global.get
```

bitbears-dev:
> The magic seems unique so i think you can add this format to the probe group, just a matter for add a `Groups: []string{format.PROBE}` to the `interp.RegisterFormat(...)` struct.

Worked like a charm 🤩


> BTW if you run `make doc` the wasm format should be added to README.md, usage.md and some other places.

done @ https://github.com/wader/fq/pull/383/commits/ce438872986fa2279748fc67de4f1a60fca23be0

bitbears-dev:
> Also formats can add their own native go function or jq functions if there are some functionallity like queries and conversions etc. For example would be interesting to try write a wasm to S-expressions function in jq? or some kind of encode support? ....or maybe even a basic interpreter?  hehe

It sounds challenging. Maybe I'll work on it after this pull request is merged. :nerd_face:

bitbears-dev:
> If you want to get nice output from fq (or other CLI tools) i wrote this thingy https://github.com/wader/ansisvg search for ansisvg in the repo for some examples.

Wow! Awesome! I'll definitely use this in my book :blue_book:

bitbears-dev:
> The `peekUnsignedLEB128` one is because you and to read a LEB128 and possibly fail but on success the position should not have changed? I'm wondering if i should generate ´Peek*` methods for all decode types?

Actually after refactoring `peekUnsignedLEB128` is no longer needed and removed. But generally speaking, it is useful to have a peek* functions.

format/wasm/mappers.go
bitbears-dev: It is from WebAssembly's specification:

image

Do you mean I should follow some godoc formatting here?

format/wasm/testdata/core/generate_fqtest.sh
bitbears-dev: Now it uses `WRITE_ACTUAL=1 go test ...` @ https://github.com/wader/fq/pull/383/commits/7401d141e87df00d916c8154b86ab5aa99cd7b0e

wader:
> > Also formats can add their own native go function or jq functions if there are some functionallity like queries and conversions etc. For example would be interesting to try write a wasm to S-expressions function in jq? or some kind of encode support? ....or maybe even a basic interpreter?  hehe
> 
> It sounds challenging. Maybe I'll work on it after this pull request is merged. 🤓

Hehe yeap let's see, i might give it shot later

wader:
> > If you want to get nice output from fq (or other CLI tools) i wrote this thingy https://github.com/wader/ansisvg search for ansisvg in the repo for some examples.
> 
> Wow! Awesome! I'll definitely use this in my book 📘

👍  also make sure to run fq with `-U` or `-o unicode=true` if you pipe the output, also there is env CLIUNICODE=1 which will use unicode if stdout is tty

wader:
> > The `peekUnsignedLEB128` one is because you and to read a LEB128 and possibly fail but on success the position should not have changed? I'm wondering if i should generate ´Peek*` methods for all decode types?
> 
> Actually after refactoring `peekUnsignedLEB128` is no longer needed and removed. But generally speaking, it is useful to have a peek* functions.

Yeap will think about how the API would look like, just additional Peek* methods of if there is some other way.

wader:
I wonder if the uses of `FieldUScalarFn` etc could be replaced with `FieldUFn`, there is also `TryFieldUFn` with a read function that can return error also, a bit simpler. Then you would have a read function that is just `func readUnsignedLEB128(d *decode.D) uint64`.

format/wasm/mappers.go
wader: Ah sorry it was about the linter complaining about godoc formatting, i like the comments with spec grammar 👍 to make the godoc formatter happy maybe they have to be verbatim somehow, maybe code "code blocks" see https://tip.golang.org/doc/comment? seem to work for me if i change:

```
// importdesc ::= 0x00 x:typeidx     => func x
//             |  0x01 tt:tabletype  => table tt
//             |  0x02 mt:memtype    => mem mt
//             |  0x03 gt:globaltype => global gt
``` 
to
```
// Some title
//
//  importdesc ::= 0x00 x:typeidx     => func x
//              |  0x01 tt:tabletype  => table tt
//              |  0x02 mt:memtype    => mem mt
//              |  0x03 gt:globaltype => global gt
```

wader:
Added LEB128 readers https://github.com/wader/fq/pull/387 did a quick ugly refactor of the wasm code to try it and seemed to work. I noticed there is `fieldI32` that does range check, so maybe you still want some kind of helper? but there are some range check helpers you can use also, something like `d.FieldULEB128("...", d.AssertURange(math.MinInt32, math.MaxInt32))` maybe? the reason the those scalar mappers are methods on `*D` is require you can disable Assert* check with `-o force=true`. `Require*` can't be disabled, used when there is no point to continue (zero length or something), `Validate*` only sets description. Should probably document this more.

format/wasm/mappers.go
bitbears-dev: Got it, I'll update comments.

wader:
Maybe put "magic" and "version " into a "header" struct? i've noticed that it is quite nice to be able to just get the header fields without enumerating them.

wader:
Maybe add some instruction constant, for the ones used when initing instrMap?

wader:
So after rebase on master to get the LEB128 and maybe the thing i commented about i think it looks ready for merge? is here something you would like to fix before i merge?

One thought i had was that the tree gets deep fast, wonder if there are possibility to flatten things? but maybe that would break following the spec? if so then let's keep it as it is i think.

Also noticed that the tests are getting a bit slow to run, but i don't think we should cut down on wasm tests. I will see if i can speed things up in other ways, reuse interpreter and maybe parallellism? also i should probably get more used to run specific tests :)

wader:
Played around with parallell tests, noticed one thing, maybe apply this patch:
```diff
+var instrMapOnce sync.Once
+
 func decodeWASM(d *decode.D, _ any) any {
        d.Endian = decode.LittleEndian

-       // delayed initialization to break initialization reference cycle
-       instrMap[0x02] = instructionInfo{mnemonic: "block", f: decodeBlock}
-       instrMap[0x03] = instructionInfo{mnemonic: "loop", f: decodeLoop}
-       instrMap[0x04] = instructionInfo{mnemonic: "if", f: decodeIf}
+       instrMapOnce.Do(func() {
+               // delayed initialization to break initialization reference cycle
+               instrMap[0x02] = instructionInfo{mnemonic: "block", f: decodeBlock}
+               instrMap[0x03] = instructionInfo{mnemonic: "loop", f: decodeLoop}
+               instrMap[0x04] = instructionInfo{mnemonic: "if", f: decodeIf}
+       })

        decodeWASMModule(d)
```

bitbears-dev:
> Played around with parallell tests, noticed one thing, maybe apply this patch:
> 
> ```diff
> +var instrMapOnce sync.Once
> +
>  func decodeWASM(d *decode.D, _ any) any {
>         d.Endian = decode.LittleEndian
> 
> -       // delayed initialization to break initialization reference cycle
> -       instrMap[0x02] = instructionInfo{mnemonic: "block", f: decodeBlock}
> -       instrMap[0x03] = instructionInfo{mnemonic: "loop", f: decodeLoop}
> -       instrMap[0x04] = instructionInfo{mnemonic: "if", f: decodeIf}
> +       instrMapOnce.Do(func() {
> +               // delayed initialization to break initialization reference cycle
> +               instrMap[0x02] = instructionInfo{mnemonic: "block", f: decodeBlock}
> +               instrMap[0x03] = instructionInfo{mnemonic: "loop", f: decodeLoop}
> +               instrMap[0x04] = instructionInfo{mnemonic: "if", f: decodeIf}
> +       })
> 
>         decodeWASMModule(d)
> ```

Oh, ok, thanks!

bitbears-dev:
> Maybe put "magic" and "version " into a "header" struct? i've noticed that it is quite nice to be able to just get the header fields without enumerating them.

There is no such header structure in the Wasm spec, magic and version are defined flat. If we introduced it, users might get confused. But I understand the point you make. If future versions of wasm add more header fields, they will need to be enumerated. It must be quite boring. :thinking:

bitbears-dev:
> Maybe add some instruction constant, for the ones used when initing instrMap?

Do you mean `0x02` `0x03` and `0x04` in `decodeWASM()` function?

bitbears-dev:
> One thought i had was that the tree gets deep fast, wonder if there are possibility to flatten things? but maybe that would break following the spec? if so then let's keep it as it is i think.

One thing that could be done is to flatten the `n` and `x` fields that are generated when decoding vectors. 

i.e. 
```
magic: raw bits (valid) 0x0-0x3.7 (4)
version: 1 0x4-0x7.7 (4)
sections[0:6]: 0x8-0x2b8.7 (689)
  [0]{}: section 0x8-0x13.7 (12)
    id: "type_section" (1) 0x8-0x8.7 (1)
    size: 10 (valid) 0x9-0x9.7 (1)
    content{}: 0xa-0x13.7 (10)
      ft{}: 0xa-0x13.7 (10)
        n: 2 (valid) 0xa-0xa.7 (1)
        x[0:2]: 0xb-0x13.7 (9)
          [0]{}: ft 0xb-0xf.7 (5)
            tag: 0x60 (valid) 0xb-0xb.7 (1)
            rt1{}: 0xc-0xd.7 (2)
              n: 1 (valid) 0xc-0xc.7 (1)
              x[0:1]: 0xd-0xd.7 (1)
                [0]: "i32" (0x7f) t 0xd-0xd.7 (1)
            rt2{}: 0xe-0xf.7 (2)
              n: 1 (valid) 0xe-0xe.7 (1)
              x[0:1]: 0xf-0xf.7 (1)
                [0]: "i32" (0x7f) t 0xf-0xf.7 (1)
          [1]{}: ft 0x10-0x13.7 (4)
            tag: 0x60 (valid) 0x10-0x10.7 (1)
            rt1{}: 0x11-0x12.7 (2)
              n: 1 (valid) 0x11-0x11.7 (1)
              x[0:1]: 0x12-0x12.7 (1)
                [0]: "i32" (0x7f) t 0x12-0x12.7 (1)
            rt2{}: 0x13-0x13.7 (1)
              n: 0 (valid) 0x13-0x13.7 (1)
              x[0:0]: 0x14-NA (0)
```

will be like:

```
magic: raw bits (valid) 0x0-0x3.7 (4)
version: 1 0x4-0x7.7 (4)
sections[0:6]: 0x8-0x2b8.7 (689)
  [0]{}: section 0x8-0x13.7 (12)
    id: "type_section" (1) 0x8-0x8.7 (1)
    size: 10 (valid) 0x9-0x9.7 (1)
    content{}: 0xa-0x13.7 (10)
      ft[0:2]: 0xa-0x13.7 (10)
        [0]{}: ft 0xb-0xf.7 (5)
          tag: 0x60 (valid) 0xb-0xb.7 (1)
          rt1[0:1]: 0xc-0xd.7 (2)
            [0]: "i32" (0x7f) t 0xd-0xd.7 (1)
          rt2[0:1]: 0xe-0xf.7 (2)
            [0]: "i32" (0x7f) t 0xf-0xf.7 (1)
        [1]{}: ft 0x10-0x13.7 (4)
          tag: 0x60 (valid) 0x10-0x10.7 (1)
          rt1[0:1]: 0x11-0x12.7 (2)
              [0]: "i32" (0x7f) t 0x12-0x12.7 (1)
          rt2[0:0]: 0x13-0x13.7 (1)
```

wader:
> > Maybe put "magic" and "version " into a "header" struct? i've noticed that it is quite nice to be able to just get the header fields without enumerating them.
> 
> There is no such header structure in the Wasm spec, magic and version are defined flat. If we introduced it, users might get confused. But I understand the point you make. If future versions of wasm add more header fields, they will need to be enumerated. It must be quite boring. 🤔

Aha good point, yes i agree let's follow spec as much as possible

wader:
> > Maybe add some instruction constant, for the ones used when initing instrMap?
> 
> Do you mean `0x02` `0x03` and `0x04` in `decodeWASM()` function?

Yeap

wader:
> > One thought i had was that the tree gets deep fast, wonder if there are possibility to flatten things? but maybe that would break following the spec? if so then let's keep it as it is i think.
> 
> One thing that could be done is to flatten the `n` and `x` fields that are generated when decoding vectors.
> 
> i.e.
> 
> ```
> magic: raw bits (valid) 0x0-0x3.7 (4)
> version: 1 0x4-0x7.7 (4)
> sections[0:6]: 0x8-0x2b8.7 (689)
>   [0]{}: section 0x8-0x13.7 (12)
>     id: "type_section" (1) 0x8-0x8.7 (1)
>     size: 10 (valid) 0x9-0x9.7 (1)
>     content{}: 0xa-0x13.7 (10)
>       ft{}: 0xa-0x13.7 (10)
>         n: 2 (valid) 0xa-0xa.7 (1)
>         x[0:2]: 0xb-0x13.7 (9)
>           [0]{}: ft 0xb-0xf.7 (5)
>             tag: 0x60 (valid) 0xb-0xb.7 (1)
>             rt1{}: 0xc-0xd.7 (2)
>               n: 1 (valid) 0xc-0xc.7 (1)
>               x[0:1]: 0xd-0xd.7 (1)
>                 [0]: "i32" (0x7f) t 0xd-0xd.7 (1)
>             rt2{}: 0xe-0xf.7 (2)
>               n: 1 (valid) 0xe-0xe.7 (1)
>               x[0:1]: 0xf-0xf.7 (1)
>                 [0]: "i32" (0x7f) t 0xf-0xf.7 (1)
>           [1]{}: ft 0x10-0x13.7 (4)
>             tag: 0x60 (valid) 0x10-0x10.7 (1)
>             rt1{}: 0x11-0x12.7 (2)
>               n: 1 (valid) 0x11-0x11.7 (1)
>               x[0:1]: 0x12-0x12.7 (1)
>                 [0]: "i32" (0x7f) t 0x12-0x12.7 (1)
>             rt2{}: 0x13-0x13.7 (1)
>               n: 0 (valid) 0x13-0x13.7 (1)
>               x[0:0]: 0x14-NA (0)
> ```
> 
> will be like:
> 
> ```
> magic: raw bits (valid) 0x0-0x3.7 (4)
> version: 1 0x4-0x7.7 (4)
> sections[0:6]: 0x8-0x2b8.7 (689)
>   [0]{}: section 0x8-0x13.7 (12)
>     id: "type_section" (1) 0x8-0x8.7 (1)
>     size: 10 (valid) 0x9-0x9.7 (1)
>     content{}: 0xa-0x13.7 (10)
>       ft[0:2]: 0xa-0x13.7 (10)
>         [0]{}: ft 0xb-0xf.7 (5)
>           tag: 0x60 (valid) 0xb-0xb.7 (1)
>           rt1[0:1]: 0xc-0xd.7 (2)
>             [0]: "i32" (0x7f) t 0xd-0xd.7 (1)
>           rt2[0:1]: 0xe-0xf.7 (2)
>             [0]: "i32" (0x7f) t 0xf-0xf.7 (1)
>         [1]{}: ft 0x10-0x13.7 (4)
>           tag: 0x60 (valid) 0x10-0x10.7 (1)
>           rt1[0:1]: 0x11-0x12.7 (2)
>               [0]: "i32" (0x7f) t 0x12-0x12.7 (1)
>           rt2[0:0]: 0x13-0x13.7 (1)
> ```

That would mean the `n` bits (length of vector?) would not be in any field? if so keep as it i think, and i guess same as with  header applies here, follow spec is best?

wader:
Test failures seems to be because of old test output

bitbears-dev:
> That would mean the `n` bits (length of vector?) would not be in any field? if so keep as it i think, and i guess same as with header applies here, follow spec is best?

Yes, `n` is number of items in vector. Vector is defined in wasm spec as following:

```
vec(B) ::= n:u32 (x:B)^n  => x^n
```

where B is type of elements, n is number of elements.

+1 to follow spec. (i.e. keep the current implementation as is)

bitbears-dev:
> github username, name, email

Perfect. Thank you.

wader:
Lint error seems godoc things again, strange. Maybe we're using different fmt versions somehow? if i save in vscode it seems to reformat (maybe needs some checking after) and lint is happy.

One possible nitpick, maybe add a comment for:
```go
...
         // set in decodeWASM
	0x02: {mnemonic: "block"},
	0x03: {mnemonic: "loop"},
	0x04: {mnemonic: "if"},
```

Test fails i think is just regenerate fqtest files again.

Otherwise i'm happy to merge if you are?

bitbears-dev:
> maybe add a comment for:
> 
> ```go
> ...
>         // set in decodeWASM
> 	0x02: {mnemonic: "block"},
> 	0x03: {mnemonic: "loop"},
> 	0x04: {mnemonic: "if"},
> ```

Good point. Added @ https://github.com/wader/fq/pull/383/commits/cd037c514d19a7d4330b4ae5bf5816784543fabc

bitbears-dev:
> Otherwise i'm happy to merge if you are?

Yes, please!

wader:
Sorry maybe was unclear, also good if CI build is green before merge, then the tests have ran on some other platforms also. Should just be to adjust some comment and a `WRITE_ACTUAL=1 go test ./format` i think?

wader:
Windows build failed with `panic: test timed out after 10m0s` hmm, i guess we can increase a bit but i will probably look into speeding this up after merge.

Can you add a `-timeout 20m` to the `make test` `go test` command line maybe?

bitbears-dev:
> Can you add a `-timeout 20m` to the `make test` `go test` command line maybe?

Done @ https://github.com/wader/fq/pull/383/commits/074c22c955813c3423d644c59272189604c53f6a

:crossed_fingers:

wader:
I noticed that some of the wasm testa are much slower than others and output was large for those i think, might be a clue. So i think we can try these things:

- Run in parallell
- Speed up output/diff or whatever is slow with large output, or is it decoding that is slow?
- Maybe not use `dv` for some test if we're not interesting in all output?
- Maybe not run all test with `-race` (is quite a lot slower)?
- Reuse interpreter somehow?

Let me know if you have more ideas

wader:
🎉

wader:
I think this should fix most of it https://github.com/wader/fq/pull/393

bitbears-dev:
Thank you for your valuable review comments and support! I'm so happy with my first contribution to `fq` :tada:

PR #385, opened 2022-08-19
interp: Add --value-output/-V option to do tovalue before output

wader:
Idea is to skip display so that JSON is outputted instead of showing tree and hexdump etc.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -40,10 +40,21 @@ fq 'd({verbose: true})' file
 
 # JSON representation for whole file
 fq tovalue file
+# or use -V (--value-output) that does tovalue automatically
+fq -V . file
+# or -Vr if the value is a string and you want a "raw" string
+fq -Vr .path.to.string file
 # JSON but raw bit fields truncated
 fq -o bits_format=truncate tovalue file
 # JSON but raw bit fields as md5 hex string
 fq -o bits_format=md5 tovalue file
+# look up a path
+fq '.some[1].path' file
+# look up a path and output JSON
+fq -V '.some[1].path' file
+# can be a query that outputs multiple values
+# this outputs first and last value in .same array and .path, three values in total
+fq -V '.some[0,-1], .path' file
 
 # grep whole tree by value
 fq 'grep("^prefix")' file
@@ -73,7 +84,7 @@ fq -o force=true -d mp4 file.mp4
 
 ### CLI arguments
 
-Most of fq's CLI argument are borrowed from jq and works in the same way.
+Most of jq's CLI arguments work with fq. But here are some additional ones specific to fq:
 
 #### Decode format `--decode`, `-d NAME`
 
@@ -85,7 +96,7 @@ Force format to decode instead of probing.
 
 Start interactive REPL.
 
-Can be used with both no, one and multiple inputs, ex just `fq -i ` start a REPL with `null` input, `fq -i 123` with the number 123 as input, `fq -i . a b` with two files as input. Also works with `--slurp`.
+Can be used with no input, one and multiple inputs, for example just `fq -i ` starts a REPL with `null` input, `fq -i 123` with the number 123 as input, `fq -i . a b` with two files as input. This also works with `--slurp`. In the REPL it is also possible to start a sub-REPLs by ending a query with `<query> | repl`, use ctrl-D to exit the sub-REPL. The sub-REPL will evaluate separately on each output from the query it was started. Use `[<query>] | repl` if you want to "slurp" into an array.
 
 #### Set option `--options`,`-o KEY=VALUE|@PATH`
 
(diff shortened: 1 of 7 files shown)


wader:
Not sure about this one yet, want something like it but is `-V` good? also should add documentation.

Demo:
```
$ fq '.headers[0].magic' format/mp3/testdata/test.mp3
   │00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 10 11│0123456789abcdef01│
0x0│49 44 33                                             │ID3               │.headers[0].magic: "ID3" (valid)

$ fq -V '.headers[0].magic' format/mp3/testdata/test.mp3
"ID3"

$ fq -Vr '.headers[0].magic' format/mp3/testdata/test.mp3
ID3

$ fq '.headers[0].flags' format/mp3/testdata/test.mp3
   │00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f 10 11│0123456789abcdef01│.headers[0].flags{}:
0x0│               00                                    │     .            │  unsynchronisation: false
0x0│               00                                    │     .            │  extended_header: false
0x0│               00                                    │     .            │  experimental_indicator: false
0x0│               00                                    │     .            │  unused: 0

$ fq -V '.headers[0].flags' format/mp3/testdata/test.mp3
{
  "experimental_indicator": false,
  "extended_header": false,
  "unsynchronisation": false,
  "unused": 0
}
```

peterwaller-arm:
👍 would have found this useful as a new user.

wader:
Hey, yeap would be nice cli ergonomic to not have to do `... | tovalue`. Any idea for something else than `-V`? also i see now that "description: "Output JSON value (don't display)" might not be very clear if you don't know that `d` etc is an alias for a `display` function.

I'll rebase the PR and fix the help description so you can try it if you want.

peterwaller-arm:
My original expectation was for -r to do something like this, which would have been the parallel to jq. But if it's not reasonable to change `-r`'s output at this point, your existing choice of UX looks reasonable to me.

wader:
I was thinking of piggyback on `-r` but it would be confusing for strings i guess? then `-r` would both make a decode value string be outputted as JSON and be raw (not quoted)

peterwaller-arm:
I'm unsure what you mean by 'do both' other than that it is a conflict and there is a choice to be made. I presume you'd want to preserve backwards compatible behaviour.

wader:
For example:
```
$ fq '.path.to.a.string' file
<fancy hexdump/tree for string>
$ fq -V '.path.to.a.string' file
"some string"
$ fq -Vr '.path.to.a.string' file
some string
```

If we used `-r` for this then the `-V` case would not be possible for strings, would always be raw string and not JSON string.

peterwaller-arm:
I think I understand what you're saying and it makes sense to me in my limited experience of fq and reasonable level of experience of jq.

wader:
Updated the documentation a bit also. Maybe the `-Vr` combination should be mentioned somewhere

wader:
Added `-Vr` to documentation and don't think i can come up with anything better than `-V`.

@peterwaller-arm  Looks good? and thanks for bumping this issue 👍 had forgot about it

doc/usage.md
```diff
@@ -73,7 +84,7 @@ fq -o force=true -d mp4 file.mp4
 
 ### CLI arguments
 
-Most of fq's CLI argument are borrowed from jq and works in the same way.
+Most of jq's CLI argument work with fq. But here are some additional ones specific to fq:
```
peterwaller-arm, line 87:
```suggestion
Most of jq's CLI arguments work with fq. But here are some additional ones specific to fq:
```

doc/usage.md
```diff
@@ -85,7 +96,7 @@ Force format to decode instead of probing.
 
 Start interactive REPL.
 
-Can be used with both no, one and multiple inputs, ex just `fq -i ` start a REPL with `null` input, `fq -i 123` with the number 123 as input, `fq -i . a b` with two files as input. Also works with `--slurp`.
+Can be used with no input, one and multiple inputs, ex just `fq -i ` start a REPL with `null` input, `fq -i 123` with the number 123 as input, `fq -i . a b` with two files as input. Also works with `--slurp`. In the REPL it is also possible to start a sub-REPL:s by ending a query with `<query> | repl`, use ctrl-D to exit the sub-REPL. The sub-REPL will evaluate separately on each output from the query it was started. Use `[<query>] | repl` if you want to "slurp" into an array.
```
peterwaller-arm, line 99:
```suggestion
Can be used with no input, one and multiple inputs, for example `fq -i ` starts a REPL with `null` input, `fq -i 123` with the number 123 as input, `fq -i . a b` with two files as input. This also works with `--slurp`. In the REPL it is also possible to start a sub-REPLs by ending a query with `<query> | repl`, use ctrl-D to exit the sub-REPL. The sub-REPL will evaluate separately on each output from the query it was started. Use `[<query>] | repl` if you want to "slurp" into an array.
```

peterwaller-arm (approved):
LGTM from inspection of the diff, some minor potential typographical improvements on the docs from a native speaker (but even the original notey-form read fine as it was).

wader:
Thanks for reviewing and sorry i modified some unrelated documentation, could not resist. The whole documentation probably needs an overhaul, parts have been written independently at different times so probably mixes styles etc.

PR #388, opened 2022-08-20
fq: Release 0.0.9

wader:


wader:
Still want to wait for wasm support and also include proper column truncate when displaying deep trees etc

wader:
@dundee @siraben @thushan @danfe @mpanczyk @orhun @herbygillot Hey, 0.0.9 have been published.

Biggest news is a `wasm` decoder otherwise mostly smaller improvement and bug fixes.

See [CHANGES.md](https://github.com/wader/fq/blob/master/CHANGES.md) for changes.

herbygillot:
Thanks @wader - MacPorts has been updated:
- https://github.com/macports/macports-ports/commit/41b343fb237ed6cc9409ca6a5573ae185f026d33
- https://ports.macports.org/port/fq/

wader:
@herbygillot Thanks! 👍

wader:
@herbygillot Noticed that there were some build failure, on the error it looks like it might be building with an old version of go  before generics support

herbygillot:
> @herbygillot Noticed that there were some build failure, on the error it looks like it might be building with an old version of go before generics support

Those are known failures, as we have seen issues trying to run Go 1.18 and newer on many older versions of macOS.  Those are still running 1.17, hence the cause of these failures.

thushan:
Confirmed working on Windows.

dundee:
0.0.9 finally available in Debian testing. Thanks for patience everyone!

wader:
@dundee Thanks! don't see adding much more module dependencies in the future, maybe simplifies debian packaging a bit?

dundee:
Yes, adding not very known dependencies (like gopacket) was the hardest part.

wader:
> Yes, adding not very known dependencies (like gopacket) was the hardest part.

👍 

Aha gopacket used to be https://github.com/google/gopacket but has been moved to its own org to easy maintenance

dundee:
Yes, it was somehow forked several times. Adding a new package to Debian is always quite time consuming job.

Issue #408, opened 2022-08-31
Increase performance


pnsafonov:
### What version are you using (`fq -v`)?

<pre>
$ fq -v
0.0.9 (linux amd64)
</pre>

### How was fq installed?

<pre>
fq is build from source.
</pre>
My branch: 
[https://github.com/pnsafonov/fq/tree/postgres](https://github.com/pnsafonov/fq/tree/postgres)

### Can you reproduce the problem using the latest release or master branch?
Yes
### What did you do?
I am implementing PostgreSQL format parsers. I can offer perfomance fix.

<pre>
time fq -d pgheap -o flavour=postgres14 ".Pages[0].PageHeaderData.pd_linp[0,-1] | tovalue" 16397_8
{
  "lp_flags": "LP_NORMAL",
  "lp_len": 121,
  "lp_off": 8064
}
{
  "lp_flags": "LP_NORMAL",
  "lp_len": 121,
  "lp_off": 384
}

real    1m20.079s
user    1m21.643s
sys     0m0.389s
</pre>

Then i made fix, and run again:
<pre>
time fq -d pgheap -o flavour=postgres14 ".Pages[0].PageHeaderData.pd_linp[0,-1] | tovalue" 16397_8
{
  "lp_flags": "LP_NORMAL",
  "lp_len": 121,
  "lp_off": 8064
}
{
  "lp_flags": "LP_NORMAL",
  "lp_len": 121,
  "lp_off": 384
}

real    0m2.076s
user    0m3.452s
sys     0m0.242s
</pre>

This fix decrease execution time 1m20.079s -> 0m2.076s, file `decode.go`:
<pre>
func (d *D) AddChild(v *Value) {
	v.Parent = d.Value

	switch fv := d.Value.V.(type) {
	case *Compound:
		//if !fv.IsArray {
		//	for _, ff := range fv.Children {
		//		if ff.Name == v.Name {
		//			d.Fatalf("%q already exist in struct %s", v.Name, d.Value.Name)
		//		}
		//	}
		//}
		fv.Children = append(fv.Children, v)
	}
}
</pre>

Also i made some perf with go tools. This was how i found fix:
<pre>
   283.22s 84.00% 84.00%    321.80s 95.44%  github.com/wader/fq/pkg/decode.(*D).AddChild
    36.65s 10.87% 94.87%     36.65s 10.87%  memeqbody
     2.42s  0.72% 95.58%      4.94s  1.47%  runtime.scanobject
[...]

PR #410, opened 2022-08-31
performance: increase perfomance when build with -tag release

pnsafonov:
Conditional compilation to fix [https://github.com/wader/fq/issues/408](https://github.com/wader/fq/issues/408)

### DIFF
--- a/pkg/build/norelease.go
+++ b/pkg/build/norelease.go
@@ -1,0 +1,7 @@
+//go:build !release
+
+package build
+
+func initReleaseOptions() {
+	CheckChildAlreadyExists = true
+}
--- a/pkg/build/options.go
+++ b/pkg/build/options.go
@@ -1,0 +1,9 @@
+package build
+
+var (
+	CheckChildAlreadyExists = true
+)
+
+func init() {
+	initReleaseOptions()
+}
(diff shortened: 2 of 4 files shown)


pkg/build/norelease.go
```diff
@@ -0,0 +1,7 @@
+//go:build !release
+
+package build
+
+func initReleaseOptions() {
+	CheckChildAlreadyExists = true
+}
```
wader, line 7: I think you can skip option.go and have this in norelease.go
```
const CheckChildAlreadyExists = true
```
and this in release.go
```
const CheckChildAlreadyExists = false
```

wader:
But did my patch earlier to use a map fix the performance issue? if so i would probably prefer that to keep things simple, generally try to avoid having options if possible

pnsafonov:
> But did my patch earlier to use a map fix the performance issue? if so i would probably prefer that to keep things simple, generally try to avoid having options if possible

 Yes, patch with map fix performance issue. Same performance as conditional compile approach.
 Merge request:
 [https://github.com/wader/fq/pull/411](https://github.com/wader/fq/pull/411)

pnsafonov:
Alternative is used:
[https://github.com/wader/fq/pull/411](https://github.com/wader/fq/pull/411)

PR #411, opened 2022-09-01
perfomance: increase performance by map usage

pnsafonov:
Perfomance fix

### DIFF
--- a/pkg/decode/decode.go
+++ b/pkg/decode/decode.go
@@ -721,11 +721,13 @@ func (d *D) AddChild(v *Value) {
 	switch fv := d.Value.V.(type) {
 	case *Compound:
 		if !fv.IsArray {
-			for _, ff := range fv.Children {
-				if ff.Name == v.Name {
-					d.Fatalf("%q already exist in struct %s", v.Name, d.Value.Name)
-				}
+			if fv.Keys == nil {
+				fv.Keys = make(map[string]struct{})
 			}
+			if _, ok := fv.Keys[v.Name]; ok {
+				d.Fatalf("%q already exist in struct %s", v.Name, d.Value.Name)
+			}
+			fv.Keys[v.Name] = struct{}{}
 		}
 		fv.Children = append(fv.Children, v)
 	}
--- a/pkg/decode/value.go
+++ b/pkg/decode/value.go
@@ -13,6 +13,7 @@ type Compound struct {
 	IsArray     bool
 	RangeSorted bool
 	Children    []*Value
+	Keys        map[string]struct{}
 	Description string
 }
 


wader:
Thanks and thanks for profiling and looking into this 👍 

One future improvement related to memory usage would be to use a `*map[string]struct{}` instead which i think would save some bytes (a pointer vs a pointer, length and capacity it think?) but the indirection would cost some performance. We can experiment with that later.

wader:
Will probably do an additional PR soon if it makes sense to change it into a `map[string]*Value` so it can also be used by `interp/decode.StructDecodeValue JQValueKey(name string) any`

pnsafonov:
> Thanks and thanks for profiling and looking into this +1
> 
> One future improvement related to memory usage would be to use a `*map[string]struct{}` instead which i think would save some bytes (a pointer vs a pointer, length and capacity it think?) but the indirection would cost some performance. We can experiment with that later.

Change map[string]*Value to *map[string]*Value will not save any byte. map is implemented via pointer.

It can be checked by:
<pre>
package main

import (
	"fmt"
	"unsafe"
)

// 8
type DataA struct {
	keys map[string]struct{} // ptr to map
}

// 8
type DataB struct {
	keys *map[string]struct{} // ptr to ptr
}

// 12
type DataC struct {
	id1 int32
	id2 int32
	id3 int32
}

// 24
type DataD struct {
	values []int // ptr to array, len, cap
}

// 8
type DataE struct {
	values *[]int // ptr to ptr (ptr to array, len, cap)
}

func main() {
	d1 := DataA{
		keys: make(map[string]struct{}),
	}
	m2 := make(map[string]struct{})
	d2 := DataB{
		keys: &m2,
	}
	d3 := DataC{}
	d4 := DataD{
		values: []int{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1},
	}
	sl5 := []int{1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1}
	d5 := DataE{
		values: &sl5,
	}

	s1 := unsafe.Sizeof(d1)
	s2 := unsafe.Sizeof(d2)
	s3 := unsafe.Sizeof(d3)
	s4 := unsafe.Sizeof(d4)
	s5 := unsafe.Sizeof(d5)

	fmt.Printf("s1 = %d\n", s1)
	fmt.Printf("s2 = %d\n", s2)
	fmt.Printf("s3 = %d\n", s3)
	fmt.Printf("s4 = %d\n", s4)
	fmt.Printf("s5 = %d\n", s5)
}
</pre>

It outputs (my machine is x64, pointer is 8 bytes):
<pre>
s1 = 8
s2 = 8
s3 = 12
s4 = 24
s5 = 8
</pre>

wader:
Aha, yes i must have confused it with how a slice work

PR #420, opened 2022-09-08
test: Skip go test with -race by default

piggynl:
This patch will allow running `make test` directly on platforms that Go's race detector doesn't support, for example, ARM and RISC-V.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 GO_BUILD_FLAGS ?= -trimpath
 GO_BUILD_LDFLAGS ?= -s -w
-GO_TEST_RACE_FLAGS ?=-race
 
 all: test fq
 
@@ -11,11 +10,18 @@ fq:
 .PHONY: test
 test: testgo testjq testcli
 
+.PHONY: test-race
+test-race: testgo-race testjq testcli
+
 .PHONY: testgo
 # figure out all go pakges with test files
 testgo: PKGS=$(shell find . -name "*_test.go" | xargs -n 1 dirname | sort | uniq)
 testgo:
-	go test -timeout 20m ${GO_TEST_RACE_FLAGS} ${VERBOSE} ${COVER} ${PKGS}
+	go test -timeout 20m ${RACE} ${VERBOSE} ${COVER} ${PKGS}
+
+.PHONY: testgo-race
+testgo-race: RACE=-race
+testgo-race: testgo
 
 .PHONY: testjq
 testjq: fq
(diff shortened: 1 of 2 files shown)


wader:
Hi, thanks. I've thought about maybe skip -race by default and instead require setting GO_TEST_RACE_FLAGS manually or have some special target `make test-race`. There are some platforms like `linux/arm64` where it seems the race detector is supported but it can fail depending on kernel mmu config, see https://github.com/wader/fq/pull/378#issuecomment-1219305498

Maybe it's enough with -race for one CI build and to make it resonable easy to enable during development?

piggynl:
Hi, I think the changes now more closely match what you described, but I'm not very sure whether this is exactly what you want. Do you have any more suggestions?

Makefile
```diff
@@ -11,11 +10,20 @@ fq:
...
 testgo: PKGS=$(shell find . -name "*_test.go" | xargs -n 1 dirname | sort | uniq)
 testgo:
-	go test -timeout 20m ${GO_TEST_RACE_FLAGS} ${VERBOSE} ${COVER} ${PKGS}
+	go test -timeout 20m ${VERBOSE} ${COVER} ${PKGS}
+
+.PHONY: testgo-race
+# figure out all go pakges with test files
+testgo-race: PKGS=$(shell find . -name "*_test.go" | xargs -n 1 dirname | sort | uniq)
+testgo-race:
+	go test -timeout 20m -race ${VERBOSE} ${COVER} ${PKGS}
```
wader, line 26: Wonder if the `testgo-race` target can be something like:
```
testgo-race: RACE=-race
testgo-race: testgo
```
Add add a `${RACE}` to `testgo`? would be nice to not repeat PKGS=.. and the `go test` command.

wader:
Like it 👍 had a comment about not repeating the go test command

wader (approved):
Will merge once CI is green, thanks! Packaging fq for something interesting?

piggynl:
Thank you for reviewing this PR!

Yes. :smiley: I'm packaging fq for Arch Linux RISC-V. If you are interested, the main page is at https://archriscv.felixc.at.

wader:
Nice! will that eventually merge into "standard" arch?

piggynl:
The long term goal is to upstream the patches in [felixonmars/archriscv-packages](https://github.com/felixonmars/archriscv-packages) as much as possible, so that riscv64 could be added to Arch Linux as an alternative architecture directly. :laughing:

wader:
👍 great, also always good to build/run in multiple environment to find bugs. Let me know if you run into something strange.

PR #427, opened 2022-09-15
Adds support for Apple Binary Plist, version 00

dgmcdona:
This adds support for decoding Apple Binary Plists. The only well documented version is `00`, and is therefore the only one supported here. I have tested this on both large and small binary plists, including ones with nested dictionaries.


format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+	elementTypeUID:              {Sym: "uid", Description: "UID"},
+	elementTypeArray:            {Sym: "array", Description: "Array"},
+	elementTypeSet:              {Sym: "set", Description: "Set"},
+	elementTypeDict:             {Sym: "dict", Description: "Dictionary"},
+}
+
+// decodes the number of bits required to store the following object
+func decodeSize(d *decode.D) uint64 {
+	n := d.FieldU4("size_bits")
+	if n != 0x0F {
```
wader, line 66: For consistency i've used lower case in number literals unless there is some specific reason

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+		return uint64(n)
+	}
+
+	d.FieldU4("large_size_marker", d.AssertU(1)) // TODO: add assertion that this is 0001
```
wader, line 70: Could write as 0b0001 if the spec if specific about base for some reason, but not important

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// get the exponent value
+	n = d.FieldU4("exponent")
+
+	// calculate the number of bytes encoding the size
+	n = uint64(math.Pow(2, float64(n)))
```
wader, line 76: Could this be `n = 1 << n` etc?

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// decode that many bytes as big endian
+	n = d.FieldUBigIntBE("size", int(n*8)).Uint64()
```
wader, line 79: Maybe some sanity check on n and/or if size can't fit in 64 bit? i guess in practice it should not happen but maybe nice if decoding something that is broken

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+		case elementTypeDate:
+			d.FieldStrFn("value", func(d *decode.D) string {
+				v := d.F64()
+				t := time.Unix(int64(v), 0)
+				return t.String()
+			})
+		case elementTypeData:
+			d.FieldStruct("value", func(d *decode.D) {
+				n := decodeSize(d)
+				d.FieldRawLen("value", int64(n))
```
wader, line 116: Possibly should be *8 as RawLen takes number of bits

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+			d.FieldStruct("value", func(d *decode.D) {
+				d.FieldStrFn("value", func(d *decode.D) string {
+					n := decodeSize(d)
+					return d.UTF8(int(n))
+				})
+			})
+		case elementTypeUnicodeString:
+			d.FieldStrFn("value", func(d *decode.D) string {
+				n := decodeSize(d)
+				return d.UTF16(int(n))
```
wader, line 128: Could possible be specific UTF16 version BE/LE? default is BE and respect BOM i think

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+				d.FieldValueNil("value")
+			case boolTrue:
+				d.FieldValueBool("value", true)
+			case boolFalse:
+				d.FieldValueBool("value", false)
+			}
+		case elementTypeInt:
+			n := decodeSize(d)
+			d.FieldSFn("value", func(d *decode.D) int64 {
+				return d.UBigIntBE(int(n)).Int64()
```
wader, line 100: To save some memory this could be conditional on n > 64. Makes me nearly think that this could be hidden in the decode API but not sure if good idea

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+			})
+		case elementTypeReal:
+			n := decodeSize(d)
+			d.FieldFFn("value", func(d *decode.D) float64 {
+				return d.FE(int(n), decode.BigEndian)
```
wader, line 105: Could be just `d.F(...)` and it will use current default endian

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+	topObjectOffset  int64
+	offsetTableStart int64
+}
+
+type plist struct {
+	T *trailer
+	O []uint64
+}
+
+func bplistDecode(d *decode.D, _ any) any {
```
wader, line 203: Is big endian default endian of this format or is it a bit mixed? default is big endian and default can be changed with `d.Endian = ...` and it will be inheried down

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+	offTblOffSize    int64
+	objRefSize       int64
+	nObjects         int64
+	topObjectOffset  int64
+	offsetTableStart int64
+}
+
+type plist struct {
+	T *trailer
+	O []uint64
```
wader, line 200: For clarity could be private (lowercase)?

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+type trailer struct {
+	offTblOffSize    int64
+	objRefSize       int64
+	nObjects         int64
+	topObjectOffset  int64
+	offsetTableStart int64
+}
+
+type plist struct {
+	T *trailer
```
wader, line 199: Can this be just `T trailer` and we can skip `new(trailer)` below?

pkg/decode/decode.go
```diff
@@ -688,11 +688,15 @@ func (d *D) trySeekAbs(pos int64, fns ...func(d *D)) (int64, error) {
 	return pos, nil
 }
 
-// SeekRel seeks relative to current bit position
+// TrySeekRel seeks relative to current bit position. If one or more functions is
+// passed, the original seek position will be restored after decoding.
```
wader, line 692: 👍

wader: Would be nice if this could be limited to just one optional function but no possible to express in go

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+
+	p := new(plist)
+	p.T = t
+
+	d.SeekAbs(t.offsetTableStart*8, func(d *decode.D) {
+		i := int64(0)
+		d.FieldArrayLoop("offset_table",
+			func() bool { return i < t.nObjects },
+			func(d *decode.D) {
+				off := d.FieldU("element", 8*int(t.offTblOffSize))
```
wader, line 231: Sanity check on offTblOffSize? will error if > 64

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+
+	d.SeekAbs(d.Len()-32*8, func(d *decode.D) {
+		d.FieldStruct("trailer", func(d *decode.D) {
+			d.FieldU40("padding") // unused
+			d.FieldS8("sort_version")
+			t.offTblOffSize = d.FieldS8("offset_table_offset_size")
+			t.objRefSize = d.FieldS8("object_reference_size")
+			t.nObjects = d.FieldS64BE("object_count")
+			t.topObjectOffset = d.FieldS64BE("top_object_offset")
+			t.offsetTableStart = d.FieldS64BE("offset_table_start")
```
wader, line 219: With different default endian BE could be skipped, assuming there is no or few LE encoded values

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+	elementTypeASCIIString:      {Sym: "ascii_string", Description: "ASCII encoded string"},
+	elementTypeUnicodeString:    {Sym: "unicode_string", Description: "Unicode string"},
+	elementTypeUID:              {Sym: "uid", Description: "UID"},
+	elementTypeArray:            {Sym: "array", Description: "Array"},
+	elementTypeSet:              {Sym: "set", Description: "Set"},
+	elementTypeDict:             {Sym: "dict", Description: "Dictionary"},
+}
+
+// decodes the number of bits required to store the following object
+func decodeSize(d *decode.D) uint64 {
```
wader, line 64: If separate parts of the size encoding is not important one can decode have own scalar decoders. Would look like this functions but skip Field* calls and then use like this `d.UFn("size", decodeSize)`

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+		default:
+			d.Errorf("unknown type marker: %d", m)
+		}
+	})
+
+	return nil
+}
+
+func decodeReference(d *decode.D, p *plist, idx uint64) {
+	d.SeekAbs(int64(p.O[idx]*8), func(d *decode.D) {
```
wader, line 185: Some error if idx does not exist in slice?

wader (commented):
Added some comments, but in general looks good i think.

Maybe you want add a `bplist.md` file? see `wasm.md` etc, also remember to embed it and run `make doc && WRITE_ACTUAL=1 go test ./format`. You might need to rebase to make it work as i re-added markdown doc support some days ago.

Also could possibly add `torepr` support, see msgpack.jq etc. I can help you with that is your not familiar with jq

wader:
Some test file with be nice also. Should be just to put a bplist in `testdata/test.bplist` and a `testdata/test.fqtest` with `$ fq dv test.bplist` and run tests with WRITE_ACTUAL and inspect if output looks sane

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+				d.FieldValueBool("value", false)
+			}
+		case elementTypeInt:
+			n := decodeSize(d)
+			d.FieldSFn("value", func(d *decode.D) int64 {
+				return d.UBigIntBE(int(n)).Int64()
+			})
+		case elementTypeReal:
+			n := decodeSize(d)
+			d.FieldFFn("value", func(d *decode.D) float64 {
```
wader, line 104: I think most of these could be just `d.FieldFE("value", int(n))` or possibly `d.FieldF("value", int(n))` is default endian is BE

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+				return d.FE(int(n), decode.BigEndian)
+			})
+		case elementTypeDate:
+			d.FieldStrFn("value", func(d *decode.D) string {
```
wader, line 108: This is 64 bit float point unix time? if it was an integer you could do`d.FieldF64("value", scalar.DescriptionActualUUnixTime)` but this seems to be float, maybe should add some helper for that?

`scalar.DescriptionActualUUnixTime` sets description so you can still get the number in jq expressions

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+			})
+		case elementTypeData:
+			d.FieldStruct("value", func(d *decode.D) {
+				n := decodeSize(d)
+				d.FieldRawLen("value", int64(n))
+			})
+		case elementTypeASCIIString:
+			d.FieldStruct("value", func(d *decode.D) {
+				d.FieldStrFn("value", func(d *decode.D) string {
+					n := decodeSize(d)
```
wader, line 121: I noticed that the range for string gets a bit weird with this, it looks like the string start 4 bits earlier then it actually do. Maybe better to move out decodeSize? some like:
```go
d := decodeSize(d)
d.FieldUTF8("value", int(n));
```
or if decode size is not decoded in parts (see earlier comments):
```go
d := d.FieldUFn("size", decodeSize(d))
d.FieldUTF8("value", int(n));
```
but then `decodeSize` needs to make sure to read the 4 "padding" bits if size_bits is < 0xf to make offset correct for the string

format/bplist/bplist.go
```diff
@@ -0,0 +1,245 @@
...
+	// calculate the number of bytes encoding the size
+	n = uint64(math.Pow(2, float64(n)))
+
+	// decode that many bytes as big endian
+	n = d.FieldUBigIntBE("size", int(n*8)).Uint64()
+	return n
+}
+
+func decodeItem(d *decode.D, p *plist) any {
+	d.FieldStruct("object", func(d *decode.D) {
```
wader, line 84: Is the object struct a bit redundant? looking a the decode tree maybe the item could decode directly into the key/value struct making the tree a bit less deep? but maybe the spec wants this structure?

wader:
A torepr function could look something like this for the current tree structure:
```jq
def _bplist_torepr:
  def _f:
    ( .object
    | if .type == "singleton" then .value
      elif .type == "int" then .value
      elif .type == "real" then .value
      elif .type == "date" then .value
      elif .type == "data" then .value.data
      elif .type == "ascii_string" then .value.value
      elif .type == "unicode_string" then .value.value
      elif .type == "uid" then .value
      elif .type == "array" then
        ( .elements
        | map(_f)
        )
      elif .type == "set" then
        ( .elements
        | map(_f)
        )
      elif .type == "dict" then
        ( .dictionary.entries
        | map({key: (.key | _f), value: (.value | _f)})
        | from_entries
        )
      else  error("unknown type: \(.type)")
      end
    );
  ( .objects
  | _f
  );
```

With this diff:
```patch
--- a/format/bplist/bplist.go
+++ b/format/bplist/bplist.go
@@ -1,6 +1,7 @@
 package bplist

 import (
+       "embed"
        "math"
        "time"

@@ -10,6 +11,9 @@ import (
        "github.com/wader/fq/pkg/scalar"
 )

+//go:embed bplist.jq
+var bplistFS embed.FS
+
 func init() {
        interp.RegisterFormat(decode.Format{
                Name:        format.BPLIST,
@@ -17,7 +21,9 @@ func init() {
                Description: "Apple Binary Property List",
                Groups:      []string{format.PROBE},
                DecodeFn:    bplistDecode,
+               Functions:   []string{"torepr"},
        })
+       interp.RegisterFS(bplistFS)
 }

 const (
```

You can do:
```sh
➜  fq git:(bplist) ✗ go run . torepr /System/Library/AssetsV2/com_apple_MobileAsset_LinguisticData/c55ebecca7037d50a22fe39315a802984688c69f.asset/AssetData/parser/config.plist
{
  "CFBundleDevelopmentRegion": "en",
  "CFBundleExecutable": "$(EXECUTABLE_NAME)",
  "CFBundleIdentifier": "com.apple.NLP",
  "CFBundleInfoDictionaryVersion": "6.0",
  "CFBundleName": "$(PRODUCT_NAME)",
  "CFBundlePackageType": "BNDL",
  "CFBundleShortVersionString": "1.0",
  "CFBundleSignature": "????",
  "CFBundleVersion": "1",
  "CanonicalRegions": {
    "de": {
...
```

And things like `$ fq -r 'torepr.CanonicalRegions | toyaml' file.bplist` etc

format/bplist/bplist.go
dgmcdona: It's BE everywhere, I'll make it implicit everywhere.

format/bplist/bplist.go
dgmcdona: For purely exploratory purposes and full transparency, a user might want to see the distinct elements of that size encoding. On the other hand it does create another level of nesting. I'll defer to your judgment on which way to go since you know what has been done elsewhere.

wader: Yes it's a bit hard know where to draw the line sometimes. But if this integer encoding is very specific to bplist i think it might make sense to decode it in parts? sometimes i've also added a "synthetic" field (has no bit range) to the struct with the value it encodes, ex see the asn1 ber decoders use of `d.FieldValue*`, that way the value is also accessible in queries.

One thing i've thought about it to add some kind of metadata per scalar value so that you can inspect what kind of encoding it uses.

format/bplist/bplist.go
dgmcdona: Added a helper function for decoding unix time from float values

format/bplist/bplist.go
dgmcdona: I'll add a synthetic `size` field for each entry to make it easier to query, and rename the `size` field in `decodeSize` to `size_bigint`.

dgmcdona:
Is there a way to have the `torepr` function display the scalar mapped value instead of the value itself? Would like to have dates represented as a timestamp instead of a floating point value, since it's not obvious how this float has to be converted.

wader:
-h formats test seems to fail, is in ./interp, should probably move it to format hmm

Maybe add a torepr test?

Will have a last look when im at a computer, but looks very good now

wader:
> Is there a way to have the `torepr` function display the scalar mapped value instead of the value itself? Would like to have dates represented as a timestamp instead of a floating point value, since it's not obvious how this float has to be converted.

torepr will use the scalars default "value" which is the symbolic value if set otherwise the actual value. But in your timestamp case a scalar mapper can also set a string description, that string is used when showing the hexdump tree thingy and can also be accessed with `todescription` (there is also `toactual`, `tosym`, and `tovalue`).

Yeah that is a bit unfortunate with the weird timestamp epoch, could it make sense to make the sym value a float unix timestamp or even more confusing? In other decoders i've kept it as numbers as it seemed nicer for queries (doing comparisons etc). Maybe it would make sense to add more time functions? jq has `strptime` but i haven't used it much myself.

There is some half-finished work in fq to make `tovalue` have an option to prefer the actual value, maybe something like that could be done to prefer description? and ideas?

wader:
Another alternative is to add an option to bplist decoder for how timestamps should be handled? maybe `-o timestamp=unix`, `-o timetamp=iso8601` etc?

format/bplist/bplist.md
```diff
@@ -0,0 +1,46 @@
...
+     |                                               |                |        key{}: 0x37-0x43.7 (13)
+0x030|                     5c                        |       \        |          type: "ascii_string" (5) (ASCII encoded string) 0x37-0x37.3 (0.4)
+0x030|                     5c                        |       \        |          size_bits: 12 0x37.4-0x37.7 (0.4)
+     |                                               |                |          size: 12 0x38-NA (0)
+0x030|                        43 46 42 75 6e 64 6c 65|        CFBundle|          value: "CFBundleName" 0x38-0x43.7 (12)
+
+...
+<snip>
+...
+```
```
wader, line 29: For other formats i've skipped having outputs in the help for now as it's also used for cli help (`-h bplist` and `help(bplist)`) to make it short and also not mess up with terminal width is small. Should probably figure out something better hmm. Maybe example should at least use less width?

format/bplist/bplist.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Get JSON representation
+```
+$ fq '. | torepr' com.apple.UIAutomation.plist
```
wader, line 33: Can be just `fq torepr com.apple.UIAutomation.plist` (`.` is the identity function).

wader:
I have a feeling it will probably also fail on some help output tests that is in formats/all.

Maybe all help output tests should be moved into individual formats? i will probably do that later on. Should probably so that a format will not affects tests outside its own testdata dirctory

wader:
An interesting future feature would be to write a `toplist` in jq, there is `toxml` :)

format/bplist/testdata/bplist.fqtest
```diff
@@ -0,0 +1,2269 @@
...
+0x00|                                    10         |            .   |          exponent: 0 0xc.4-0xc.7 (0.4)
+0x00|                                       13      |             .  |          size_bigint: 19 0xd-0xd.7 (1)
+    |                                               |                |          size: 19 0xe-NA (0)
+0x00|                                          55 49|              UI|          value: "UIAutomationEnabled" 0xe-0x20.7 (19)
+0x10|41 75 74 6f 6d 61 74 69 6f 6e 45 6e 61 62 6c 65|AutomationEnable|
+0x20|64                                             |d               |
+    |                                               |                |        value{}: 0x21-0x21.7 (1)
+0x20|   09                                          | .              |          type: "singleton" (0) (Singleton value (null/bool)) 0x21-0x21.3 (0.4)
+    |                                               |                |          value: true 0x22-NA (0)
+0x20|   09                                          | .              |  unknown0: raw bits 0x21.4-0x21.7 (0.4)
```
wader, line 27: I think these are the unused 4 bit when size is < 0xf? maybe add a "unused" field for them etc?

format/bplist/bplist.jq
```diff
@@ -0,0 +1,30 @@
...
+        ( .entries
+        | map({key: (.key | _f), value: (.value | _f)})
+        | from_entries
+        )
+      else  error("unknown type: \(.type)")
+      end
+     );
+  ( .objects
+  | _f
+   );
```
wader, line 29: Wops weird indent

format/bplist/bplist.go
```diff
@@ -0,0 +1,246 @@
...
+	d.FieldStruct("header", func(d *decode.D) {
+		d.FieldUTF8("magic", 6, d.AssertStr("bplist"))
+		d.FieldUTF8("version", 2, d.AssertStr("00"))
+	})
+
+	p := new(plist)
+
+	d.SeekAbs(d.Len()-32*8, func(d *decode.D) {
+		d.FieldStruct("trailer", func(d *decode.D) {
+			d.FieldU40("padding") // unused
```
wader, line 217: Make the field name unused?

format/bplist/bplist.go
```diff
@@ -0,0 +1,246 @@
...
+			func(d *decode.D) {
+				idx := d.FieldU8("object_index")
+				decodeReference(d, p, idx)
+				i++
+			})
+	case elementTypeSet:
+		n := decodeSize(d)
+		d.FieldValueU("size", n)
+		i := uint64(0)
+		d.FieldStructArrayLoop("entries", "entry",
```
wader, line 150: In some other decoders i've skipped using the `d.FieldStrctArrayLoops()` and instead use `for` as I think it feels a bit awkward. Any idea how to make it nicer? maybe a specialized version for given number of entries? maybe with generics it can be made nicer?

wader: But i'm fine with this also, it's easy to find and possibly refactor in the future

wader:
Hey, i made some changes to help and help tests https://github.com/wader/fq/pull/430 for you i think it's just to run `make doc` again and possible add a help test, see some help_*.fqtest file

dgmcdona:
Thanks! I've been super busy but hopefully I can put the finishing touches on this this weekend.

wader:
No worries, great, yes i think it's just a few small things left. But i will take an extra look if there is anything so that you have all comments and suggestions ready for the weekend

format/all/help.fqtest
```diff
@@ -293,6 +293,16 @@ out   $ fq -d bitcoin_transaction . file
...
+"help(bplist)"
+out bplist: Apple Binary Property List decoder
+out 
+out # Decode examples
+out 
+out   # Decode file as bplist
+out   $ fq -d bplist . file
+out   # Decode value as bplist
+out   ... | bplist
+out
```
wader, line 305: This file is gone now if it causes confusing conflict

format/bplist/bplist.go
```diff
@@ -0,0 +1,246 @@
...
+			},
+		)
+	})
+
+	d.SeekAbs(int64(p.o[0] * 8))
+
+	d.FieldStruct("objects",
+		func(d *decode.D) {
+			decodeItem(d, p)
+		})
```
wader, line 244: Could this somehow be `decodeReference(d, p, 0)`? looks similar but maybe confusing?

dgmcdona: I think that works, I prefer it over the seek operation.

format/bplist/bplist.go
dgmcdona: Added a function `FieldStructNArray` for decoding arrays with a known number of elements, does that look okay?

format/bplist/testdata/bplist.fqtest
dgmcdona: I'm not seeing this in the output anymore, so I think it is fixed? Can't remember right now if I explicitly fixed this, let me know if not.

dgmcdona:
Is `FieldRawLen` the correct way to decode a blob of binary data, such as in the `Data` type for binary plists? I'm having trouble extracting the data using `toactual` in the case that I might want to write it out into another file. (although I may just be missing something).

wader:
> Is `FieldRawLen` the correct way to decode a blob of binary data, such as in the `Data` type for binary plists? I'm having trouble extracting the data using `toactual` in the case that I might want to write it out into another file (although I may just be missing something). I get an error like:
> 
> ```
> can't convert actual value jq value &bitio.SectionReader{r:(*bitio.SectionReader)(0xc0009a53b0), bitBase:768, bitOff:768, bitLimit:128128}
> ```
> 
> Also, it seems weird that `tovalue` produces a base64 encoding, but it is truncated for longer values:
> 
> ```
> ./fq 'torepr.SandboxProfileData | tovalue' /Users/davidmcdonald/Library/Containers/com.apple.LoginUserService/Container.plist
> 
> "<15920>AACgAJIAAAAFAAAALgASAV8DLAMWBJ8AnQCcAJ8AnwCfAJ8AngCfAJ4AnwCbAJ4AlACDAH4AfQCbAHwAfABhAF4AngBcAJsAmwCeAHwAUgBSAEwASQBSAFIAUgBSAJ8AUgBDAEAAnwCfAJ8AnwCeAJ8AnwCCAJ8AnwCfADsANQCeAJ8AnwCfAJ8AnwCfAJ8AnwCfADQAMAA0AC8ALAArAJ8AnwCfAJ8AnwCfAJ4AnwCeAJ8AnwCfAJ8AHwCfAJ8AnwAcAJ4AGwAbABoAFQCfAJ8AFACfAJ8AnwATABMAEwAOAA4AngCfAJ4AEwCfABMAngATABMAEwCeAAwAngALAA=="
> ```
> 
> In this case, only 0x100 of the 15920 bytes are encoded as base64, the rest is missing

Yes raw is for raw bits (also does not have to be even bytes). Short version is to use `tobits`/`tobytes` (depending on which unit size for slicing you want) to get raw bits, otherwise it will behave more as "preview" string. See longer version below.

So at the moment you can do something like this:

```sh
# will show hexdump if stdout is a tty (to be safe), can do ... | cat if you really want raw bytes in the tty
$ go run . 'torepr.SandboxProfileData | tobytes' /Users/wader//Library/Containers/com.apple.LoginUserService/Container.plist

# will write raw data as stdout
$ go run . 'torepr.SandboxProfileData | tobytes' /Users/wader//Library/Containers/com.apple.LoginUserService/Container.plist > data
```

Longer version:

Problem comes from how to represent binary data as jq values. I've tried a couple of different variants, introduce a new `binary` type, array of ints and as strings, they all have different drawbacks and issues.

- A new `binary` type feels natural to add but turns out the jq standard library code etc has assumptions about which types can exist. In fq you can still get the "real" type using `_exttype`, `[0xff,0x90] | tobytes | _exttype` is "binary".  Also it still have to be able to be a JSON compatible value at some point.
- Array of ints would still have to be special in some way to know how big each element is (bit/byte).
- Strings in jq are unicode codepoints arrays so binary data might be interpreted as multi-byte codepoints and i think there were other confusing issue making not suitable for binary, see ex: `"åäö" | .,tobytes | length`.

Also there is the issue what to do with some formats that has raw fields that can be very large (ex: mp4 mdat), include all or truncates somehow?

So the current compromise is that raw will be base64 string (to be jq compatible) and also be truncated by default. It is possible change the `tovalue` truncate behaviour with `-o bits_format=base64` (there is also `md5`).

Sorry for the long rant :) but it's very good someone else is messing around with this as i'm not that happy with the current design and i think it can be made better and less confusing, so feedback is very welcome.

BTW fq has support for "binary arrays" (similar to iolists in erlang) so you slice and concatenate parts into a new binary. Maybe a not very good example:
```sh
# build a binary array with a bytes (0), a binary slice and a string (will be utf8 bytes) and try to decode it as s bplist (force to skip magic check).
$ go run . -n '"hello" | tobits | [0, .[8:16], "a string"] | tobytes | bplist({force: true}) | d'
   │00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f│0123456789abcdef│.{}: (bplist)
   │                                               │                │  error: bplist: SeekAbs: failed at position 8 (read size 0 seek pos 0): invalid seek offset
   │                                               │                │  header{}:
0x0│00 65 61 20 73 74                              │.ea st          │    magic: "\x00ea st" (invalid)
0x0│                  72 69                        │      ri        │    version: "ri" (invalid)
0x0│                        6e 67│                 │        ng│     │  unknown0: raw bits
```

dgmcdona:
Handling timestamps is tricky. On the one hand, it's tempting to render it as the description by default, since that is the behavior of `plutil` and presents the value clearly to the user. On the other hand, `jq` does have functions for converting timestamps between formats from the raw value. I think the best solution for now is to be true to the original format and render the value as the decoded floating point value, and access the description using `todescription` as you suggested, that way we stay consistent with the `jq` way of doing things. I've added a note on this in the format documentation.

wader:
singletons seems to produce unknown fields:
```sh
➜  fq git:(bplist) ✗ go run . -o line_bytes=10 'grep_by(.type=="singleton"), .unknown0, .unknown1 | dv' /Users/wader//Library/Containers/com.apple.LoginUserService/Container.plist
      │00 01 02 03 04 05 06 07 08 09│0123456789│.objects.entries[3].value.entries[5].value.entries[0].value{}: 0x475b-0x475b.7 (1)
0x4754│                     09      │       .  │  type: "singleton" (0) (Singleton value (null/bool)) 0x475b-0x475b.3 (0.4)
      │                             │          │  value: true 0x475c-NA (0)
      │00 01 02 03 04 05 06 07 08 09│0123456789│.objects.entries[3].value.entries[5].value.entries[1].value{}: 0x475b-0x475b.7 (1)
0x4754│                     09      │       .  │  type: "singleton" (0) (Singleton value (null/bool)) 0x475b-0x475b.3 (0.4)
      │                             │          │  value: true 0x475c-NA (0)
      │00 01 02 03 04 05 06 07 08 09│0123456789│
0x4754│                     09      │       .  │.unknown0: raw bits 0x475b.4-0x475b.7 (0.4)
      │00 01 02 03 04 05 06 07 08 09│0123456789│
0x4754│                        09   │        . │.unknown1: raw bits 0x475c-0x475c.7 (1)
```
Looks like 12 bits after the type is seen as unknown (should be one unknown field but seems there is a bug in the gap code, maybe because of synthetic fields, will check).

Also was a bit confused first until i realized that bplist can apparently use the index multiple times, quite cool.

wader:
Range gap issue fixed in master https://github.com/wader/fq/pull/431, now unknown field shows up as one 12 bit field.

But i got a bit unsure if the gap we're seeing in the example above is correct or not? does "normal" bplist have them? i guess based on how bplist uses offsets tables it would be possible to have ranges that are unused/unknown but does that happen in practice? maybe because alignment etc?

Generally i've tried to make decoders behave so that they ends up with gaps only for things they don't know about/should not be there, ex unknown trailing data.

wader:
With https://github.com/wader/fq/pull/432 (not in master yet)`toactual` and `tosym` behave the same as `tovalue`. Also fixed the error. I think that make sense?

The code that handles this is starting to get a bit out of hand, badly needs a rethink/refactor

wader:
Im thinking about releasing 0.0.10 soonish, would be nice to include this. If your busy we can merge and i can try fix the remaning things if you like?

dgmcdona:
> Im thinking about releasing 0.0.10 soonish, would be nice to include this. If your busy we can merge and i can try fix the remaning things if you like?

Sorry, I've been drowning in work and thesis so I haven't had time to figure out those last unknown bytes. If you want to merge it for the next release, I'm happy to figure out the bugs when I get a chance.

wader:
> Sorry, I've been drowning in work and thesis so I haven't had time to figure out those last unknown bytes. If you want to merge it for the next release, I'm happy to figure out the bugs when I get a chance, or you can if you have the time.

No need to be sorry, focus on thesis! what is it about?

Ok i'll let you know here if i figure something out

wader:
Hey, this should fix the unknown field for singletons:
```patch
diff --git a/format/bplist/bplist.go b/format/bplist/bplist.go
index 7c74848c..83480d0d 100644
--- a/format/bplist/bplist.go
+++ b/format/bplist/bplist.go
@@ -91,15 +91,11 @@ func decodeItem(d *decode.D, p *plist) {
        m := d.FieldU4("type", elementTypeMap)
        switch m {
        case elementTypeNullOrBoolOrFill:
-               t := d.U4()
-               switch t {
-               case null:
-                       d.FieldValueNil("value")
-               case boolTrue:
-                       d.FieldValueBool("value", true)
-               case boolFalse:
-                       d.FieldValueBool("value", false)
-               }
+               d.FieldU4("value", scalar.UToScalar{
+                       null:      scalar.S{Sym: nil},
+                       boolTrue:  scalar.S{Sym: true},
+                       boolFalse: scalar.S{Sym: false},
+               })
        case elementTypeInt:
                n := d.FieldUFn("size", func(d *decode.D) uint64 {
                        return 1 << d.U4()
```

format/bplist/bplist.md
```diff
@@ -0,0 +1,37 @@
+### Show full decoding
+```sh
+$ fq -d bplist dv Info.plist
```
wader, line 3: Maybe use `d` instead of `dv` in example?

wader: could also skip  `-d bplist` i guess?

wader:
Maybe left is to add some `torepr` test and just regenerate documentation and test actual output.

After that done i think or something more you want to do?

wader:
Let's merge and i fix the things in master

wader:
Thanks a lot for your contribution 🥳 Hope the decode API was ok to work with and that you might want to add more formats etc in the future!

wader:
Hmm noticed now that the bplist commits don't have your github email. Feel free to do some dummy PR if you want to show up as a contributor to the project.

dgmcdona:
Thanks for taking this one to the finish line! I don't like leaving things hanging in an incomplete state, it's just been a long couple of weeks. Looking forward to making some more contributions, whether it's to documentation, the API, or new format decoders. I work in the computer forensics field and would love to see this support all kinds of CF-related formats. No need to worry about doing a dummy PR right now, I'll get it fixed up for the next feature PR 😄

wader:
@dgmcdona Hey, got your message but can't see it here? strange.

Anyways, no problem and totally understand! was so little left and wanted it part of 0.0.10 :) yes i'm looking forward to future contributions and feel free to email me or open issues if you have ideas or want to discuss something, ex how fq could be used in forensics.

What kind of formats are common CF-formats? filesystems etc? I haven't used fq for that but it is designed to handle broken files and i try divide formats into smaller "subformats" to make it possible to decode parts separately etc. For example i've used fq quite a lot to search for patterns then try to decode and filter something at each match, as jq is a generator based language it is quite ergonomic to do, ex try to decode each occurrence for 0xfff8 as a FLAC frame:
```jq
tobytes as $b | scan([0xff,0xf8]) | $b[.start:] | flac_frame
````
Would be really great of someone else experimented with things like that as the functions and behaviors of the binary type in fq is a bit strange at times and i don't really know myself how i would like it to work :)

PR #451, opened 2022-10-07
fixes calculation of floating point lengths

dgmcdona:
Minor bug fix to calculations of floating point value byte count.


wader:
👍 thanks!

PR #454, opened 2022-10-07
fixes reference calculation to use reference size from trailer

dgmcdona:
Fixes code where references were always interpreted as a single byte, instead of the reference size as defined in the offset table. Also adds some sanity checking for dictionary key type, and offsets in invalid ranges.


format/bplist/testdata/bplist.fqtest
```diff
@@ -2445,3 +2445,2463 @@ $ fq dv float.plist
 0x20|      00 00 00 00 00 00 00 00                  |  ........      |    top_object_offset: 0 0x22-0x29.7 (8)
 0x20|                              00 00 00 00 00 00|          ......|    offset_table_start: 17 0x2a-0x31.7 (8)
 0x30|00 11|                                         |..|             |
+$ fq dv dicttest_broken.plist
```
wader, line 2448: Should it be renamed to dicttest_many_elements etc?

format/bplist/bplist.go
```diff
@@ -191,8 +210,26 @@ type trailer struct {
...
+	pl.objectStack = pl.objectStack[:len(pl.objectStack)-1]
+}
+
+func (pl *plist) indexIsInStack(idx uint64) bool {
+	for _, existing := range pl.objectStack {
+		if existing == idx {
+			return true
+		}
+	}
+	return false
```
wader, line 232: Nice 👍  i should probably do something like this in other places were i just have some kind of depth limit instead

format/bplist/bplist.go
```diff
@@ -104,82 +104,101 @@ func decodeItem(d *decode.D, p *plist) {
...
-		d.Errorf("index %d out of bounds for object table size %d", idx, len(p.o))
-		return
+		d.Errorf("index %d out of bounds for object table size %d", idx, len(pl.o))
+		return nil
 	}
 
-	d.SeekAbs(int64(p.o[idx]*8), func(d *decode.D) {
-		decodeItem(d, p)
+	if pl.indexIsInStack(idx) {
+		d.Errorf("recursion detected: object %d already decoded in stack %v", idx, pl.objectStack)
```
wader, line 184: This should probably be `d.Fatalf`, difference is that `d.Errorf` is ignore when using `-o force=true`. It's a way to skip magic checks and other things that wont lead to non-recoverable errors or loop etc

wader:
If not too complicated maybe add a infinite loop error test?

wader:
Great stuff 👍

format/bplist/bplist.go
wader: Maybe Errorf should be renamed to make it more clear? Assert? i think error/fatal comes from how go test framework works

dgmcdona: I think `Errorf` is okay for now, feels natural to keep it in line with `go test` syntax

format/bplist/bplist.go
dgmcdona: I though about using a map instead of a slice for looking up the previously decoded references but figured a slice would be faster for the common case where things aren't nested too deeply

wader: Yeap slice is probably fine. I've noticed that fq currently allocates quite a lot (each field is a couple allocs at least i think) so that is usually what shows up in both cpu and memory profiles reports.

dgmcdona:
Hold off on merging, need to fix an integer with signed integers

format/bplist/bplist.go
```diff
@@ -103,83 +103,113 @@ func decodeItem(d *decode.D, p *plist) {
...
-				idx := d.FieldU8("object_index")
-				decodeReference(d, p, idx)
+				idx := d.FieldU("object_index", int(p.t.objRefSize)*8)
+				p.decodeReference(d, idx)
 			})
 	case elementTypeDict:
 		n := decodeSize(d)
 		d.FieldValueU("size", n)
-		d.FieldStructNArray("entries", "entry", int64(n),
+		return d.FieldStructNArray("entries", "entry", int64(n),
```
wader, line 164: For the `FieldStructNArray` cases does it make senes to return to *D? is the return value used for dictionary key string? if so an alternative that make be clearer is to have `decodeItem` return `(string, bool)` or  just `string` if empty string is not allowed as key?

wader:
So maybe fix the key string thingy if you agree otherwise i'm ok to merge if you're happy. We can add more tests later. I've usually tried to add tests for "regressions" if possible.

BTW there is some fuzzing support if you do `make fuzz`, currently it will fuzz each format in round-robin but maybe should make it possible to fuzz just one format? see comment above the fuzz target in the Makefile for how to use it if your interested.

format/bplist/bplist.go
dgmcdona: Good call, going to return a bool indicating whether or not decoded item is a string. Still feels a little awkward but it's better than returning everything and doing type checking. I also verified that empty strings are allowed as dictionary keys.

wader: Ah yes it's not even needed to return the string, just decode in the struct? like it. There is some code in fq similar to how it looked in the previous version, would be nice to figure of some decode API for making it nicer Maybe possible with generics but one problem is that with go generics you can't add additional type parameters to a method i think? :(

wader:
LGTM, let me know if your ready to merge

dgmcdona:
I think it's ready 👍

wader:
👍  Thanks!

PR #459, opened 2022-10-11
decode,fuzz: Signed integer (S) read require at least one bit

wader:
Otherwise there is no sign bit and it will panic on shifting.

Maybe should require 2? now -0 and +1 hmm


wader:
@ktmf01 Hi, i noticed while looking into this that the flac spec says this about escaped samples:

> Note that it is possible that the number of bits is 0, which means all residual samples in that partition have a value of 0, and no bits code for the partition itself.

Do you happen to have an example file doing this?

wader:
Great, thanks!

PR #493, opened 2022-11-27
Apple bookmarkData

dgmcdona:
This PR implements a decoder for macOS/iOS `bookmarkData` blobs, which are often found within binary plist files. These are used to resolve URL objects for a file, even if the user moves or renames it. 

There are some issues that I could use help ironing out, namely getting it to work properly as a nested format using `grep_by` or `select` in recursive `jq` expressions. I think there may be something wrong with the way `torepr` and `tobytes` are working in the `bplist` decoder. For instance this does not work in the way that I would think it would (does not yield any results): 
```
fq '.. | select(format=="bookmark") | .map(. | torepr)' com.apple.LSSharedFileList.RecentApplications.sfl2
```

In this case, when a `bplist` `data` type is passed to `tobytes`, the bytes output are the truncated base64 representation: 
```
./fq 'torepr."$objects"[15] | tobytes | bookmark' ~/Library/Application\ Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.RecentApplications.sfl2
```


wader:
> This PR implements a decoder for macOS/iOS `bookmarkData` blobs, which are often found within binary plist files. These are used to resolve URL objects for a file, even if the user moves or renames it.

Nice!

> There are some issues that I could use help ironing out, namely getting it to work properly as a nested format using `grep_by` or `select` in recursive `jq` expressions. I think there may be something wrong with the way `torepr` and `tobytes` are working in the `bplist` decoder. For instance this does not work in the way that I would think it would (does not yield any results):
> 
> ```
> fq '.. | select(format=="bookmark") | .map(. | torepr)' com.apple.LSSharedFileList.RecentApplications.sfl2
> ```

Aha yes `format` is fq specific function that checks if a value is a decode "root" and if so return a string of the name of the decode, so in short `grep_by(format=="jpeg"))` would find all nested jpeg root values.

Had a quick look, maybe something like this would work?
```
fq 'grep_by(.type=="data" and .value[0:4] == "book") | bookmark | torepr' ...
```

> 
> In this case, when a `bplist` `data` type is passed to `tobytes`, the bytes output are the truncated base64 representation:
> 
> ```
> ./fq 'torepr."$objects"[15] | tobytes | bookmark' ~/Library/Application\ Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.RecentApplications.sfl2
> ```

I think the problem might be that by default fq currently turns "raw" fields into truncated "<length><base64>" strings when represented as a jq value, jq has no binary type unfortunately. Maybe have to rethink that and also i think maybe gojq had some changes that might have made strings "binary safe".

But there is an option to change how raw fields should be represented, try `fq -o bits_format=string torepr ...`, valid values are md5, base64, truncate, string and snippet (defualt). Maybe bits_format shoud be renamed to raw_format?

Nice work, i will review more tomorrow!

format/bookmark/bookmark.md
```diff
@@ -0,0 +1,42 @@
...
+  [Sarah Edwards'](https://www.mac4n6.com/blog/2016/1/1/manual-analysis-of-nskeyedarchiver-formatted-plist-files-a-review-of-the-new-os-x-1011-recent-items)
+  excellent research on the subject.
+
+Locating `bookmarkData` objects in plist files, especially those of the
+`NSKeyedArchiver` type, is a place where `fq`'s recursive searching really
+shines. Here, we recursively descend through the decoded tree, probing for and 
+selecting any `bookmark` blobs, then converting them to readable JSON with
+`torepr`:
+```
+fq '.. | select(format=="bookmark") | .map(. | torepr)' com.apple.LSSharedFileList.RecentApplications.sfl2
```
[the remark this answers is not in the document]
dgmcdona: I've noticed the exact same thing and I'm not sure what to make of it yet. I'll do some digging and see if I can understand the problem.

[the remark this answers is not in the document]
dgmcdona: > This example looks a bit strange, is it suppose to work with the current code? looks like it decodes `com.apple.LSSharedFileList.RecentApplications.sfl2` as bplist and recurses looking for things decoded as `bookmark`, is the plan to make bplist decoder somehow detect and decode booktmark?

Yeah, was looking for a quick one-liner for recovering bookmarks from any given `plist`/`sfl2` file, since that would be a big use case for forensic analysts.

format/bookmark/bookmark.md
```diff
@@ -0,0 +1,42 @@
...
+  actually `plist` files of the `NSKeyedArchiver` format. They can be parsed the
+  same as `plist` files, but they have a more complicated tree-like structure
+  than would typically be found, which can make locating and retrieving specific
+  values difficult, even once it has been converted to a JSON representation.
+  For more information about these types of files, see 
+  [Sarah Edwards'](https://www.mac4n6.com/blog/2016/1/1/manual-analysis-of-nskeyedarchiver-formatted-plist-files-a-review-of-the-new-os-x-1011-recent-items)
+  excellent research on the subject.
+
+Locating `bookmarkData` objects in plist files, especially those of the
+`NSKeyedArchiver` type, is a place where `fq`'s recursive searching really
```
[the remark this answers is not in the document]
dgmcdona: Yeah, the underlying decoding is `bplist`. As you noticed, there seem to be a lot of unknown bit ranges which I suspect is due to one or more `bplist` decoder bugs.

format/bookmark/bookmark.md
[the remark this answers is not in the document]
wader: > Yeah, was looking for a quick one-liner for recovering bookmarks from any given `plist`/`sfl2` file, since that would be a big use case for forensic analysts.

I see, make sense, should be possible. Would like the example to output one big JSON with all bookmark blob expanded "in place" in the sfl2 tree or output each bookmark blob as a separate JSON fragment? or something else?

wader:
Had look at binary safe strings in jq and gojq. jq is not safe (replaces invalid code points with "REPLACEMENT CHARACTER") but gojq is. Seems i remembered it wrong, i think maybe i confused with gojq removed support for \x## escapes.

```
$ fq -n '[range(256)] | tobytes' | gojq -jrRs . | hexdump -C
00000000  00 01 02 03 04 05 06 07  08 09 0a 0b 0c 0d 0e 0f  |................|
00000010  10 11 12 13 14 15 16 17  18 19 1a 1b 1c 1d 1e 1f  |................|
00000020  20 21 22 23 24 25 26 27  28 29 2a 2b 2c 2d 2e 2f  | !"#$%&'()*+,-./|
00000030  30 31 32 33 34 35 36 37  38 39 3a 3b 3c 3d 3e 3f  |0123456789:;<=>?|
00000040  40 41 42 43 44 45 46 47  48 49 4a 4b 4c 4d 4e 4f  |@ABCDEFGHIJKLMNO|
00000050  50 51 52 53 54 55 56 57  58 59 5a 5b 5c 5d 5e 5f  |PQRSTUVWXYZ[\]^_|
00000060  60 61 62 63 64 65 66 67  68 69 6a 6b 6c 6d 6e 6f  |`abcdefghijklmno|
00000070  70 71 72 73 74 75 76 77  78 79 7a 7b 7c 7d 7e 7f  |pqrstuvwxyz{|}~.|
00000080  80 81 82 83 84 85 86 87  88 89 8a 8b 8c 8d 8e 8f  |................|
00000090  90 91 92 93 94 95 96 97  98 99 9a 9b 9c 9d 9e 9f  |................|
000000a0  a0 a1 a2 a3 a4 a5 a6 a7  a8 a9 aa ab ac ad ae af  |................|
000000b0  b0 b1 b2 b3 b4 b5 b6 b7  b8 b9 ba bb bc bd be bf  |................|
000000c0  c0 c1 c2 c3 c4 c5 c6 c7  c8 c9 ca cb cc cd ce cf  |................|
000000d0  d0 d1 d2 d3 d4 d5 d6 d7  d8 d9 da db dc dd de df  |................|
000000e0  e0 e1 e2 e3 e4 e5 e6 e7  e8 e9 ea eb ec ed ee ef  |................|
000000f0  f0 f1 f2 f3 f4 f5 f6 f7  f8 f9 fa fb fc fd fe ff  |................|
00000100
$ fq -n '[range(256)] | tobytes' | jq -jrRs . | hexdump -C
00000000  00 01 02 03 04 05 06 07  08 09 0a 0b 0c 0d 0e 0f  |................|
00000010  10 11 12 13 14 15 16 17  18 19 1a 1b 1c 1d 1e 1f  |................|
00000020  20 21 22 23 24 25 26 27  28 29 2a 2b 2c 2d 2e 2f  | !"#$%&'()*+,-./|
00000030  30 31 32 33 34 35 36 37  38 39 3a 3b 3c 3d 3e 3f  |0123456789:;<=>?|
00000040  40 41 42 43 44 45 46 47  48 49 4a 4b 4c 4d 4e 4f  |@ABCDEFGHIJKLMNO|
00000050  50 51 52 53 54 55 56 57  58 59 5a 5b 5c 5d 5e 5f  |PQRSTUVWXYZ[\]^_|
00000060  60 61 62 63 64 65 66 67  68 69 6a 6b 6c 6d 6e 6f  |`abcdefghijklmno|
00000070  70 71 72 73 74 75 76 77  78 79 7a 7b 7c 7d 7e 7f  |pqrstuvwxyz{|}~.|
00000080  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000090  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000000a0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000000b0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000000c0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000000d0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000000e0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000000f0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000100  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000110  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000120  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000130  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000140  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000150  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000160  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000170  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000180  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000190  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000001a0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000001b0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000001c0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000001d0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000001e0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000001f0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000200
```

But maybe as expected in gojq the bytes will not survive a JSON run-trip, could maybe be worked around in fq but then would not be standard compliant i guess.

```
fq -n '[range(256)] | tobytes' | gojq -jrRs 'tojson | fromjson' | hexdump -C
00000000  00 01 02 03 04 05 06 07  08 09 0a 0b 0c 0d 0e 0f  |................|
00000010  10 11 12 13 14 15 16 17  18 19 1a 1b 1c 1d 1e 1f  |................|
00000020  20 21 22 23 24 25 26 27  28 29 2a 2b 2c 2d 2e 2f  | !"#$%&'()*+,-./|
00000030  30 31 32 33 34 35 36 37  38 39 3a 3b 3c 3d 3e 3f  |0123456789:;<=>?|
00000040  40 41 42 43 44 45 46 47  48 49 4a 4b 4c 4d 4e 4f  |@ABCDEFGHIJKLMNO|
00000050  50 51 52 53 54 55 56 57  58 59 5a 5b 5c 5d 5e 5f  |PQRSTUVWXYZ[\]^_|
00000060  60 61 62 63 64 65 66 67  68 69 6a 6b 6c 6d 6e 6f  |`abcdefghijklmno|
00000070  70 71 72 73 74 75 76 77  78 79 7a 7b 7c 7d 7e 7f  |pqrstuvwxyz{|}~.|
00000080  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000090  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000000a0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000000b0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000000c0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000000d0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000000e0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000000f0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000100  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000110  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000120  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000130  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000140  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000150  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000160  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000170  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
00000180  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
00000190  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000001a0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000001b0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000001c0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
000001d0  ef bf bd ef bf bd ef bf  bd ef bf bd ef bf bd ef  |................|
000001e0  bf bd ef bf bd ef bf bd  ef bf bd ef bf bd ef bf  |................|
000001f0  bd ef bf bd ef bf bd ef  bf bd ef bf bd ef bf bd  |................|
00000200
```

Also gojq uses go `==` operator to compare string and the golang spec says `String values are comparable and ordered, lexically byte-wise` so that should be fine also.

So i guess the only reason to still truncate would be that some formats would in common cases produce very big strings for some raw fields (ex mdat in mp4) when convert to JSON. Note that even with no truncate and use string instead of base64 torepr as JSON output would still produce JSON that would not be keeping some bytes (to JSON run-trip issue), it's only binary safe "internally" in fq, ex when use a query to find bookmark blob strings and decode them.

format/applebookmark/apple_bookmark.jq
```diff
@@ -0,0 +1,34 @@
...
+        )
+      elif .type == "Dictionary" then
+        ( .data
+        | map({key: (.key | _f), value: (.value | _f)})
+        | from_entries
+        )
+      elif .type == "UUID" then .data | tovalue
+      elif .type == "URL" then .data | tovalue
+      elif .type == "RelativeURL" then
+		.data | map(.record.data)
```
wader, line 26: Some tab character sneak in here i think. 2 space per indent seems to work nice with jq's syntax

format/applebookmark/apple_bookmark.jq
```diff
@@ -0,0 +1,34 @@
...
+      end
+     );
```
wader, line 28: One to the left to align with (, probably from bplist.jq? feel free to fix that one also

dgmcdona:
You're right, UID was reading short. I fixed that so that it reads correctly now. I'm still seeing bitfields unaccounted for, but a side-by-side comparison of the current `torepr` display and the output of `plutil -p` doesn't show any discrepancies or missing data.

format/bookmark/bookmark.md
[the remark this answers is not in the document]
dgmcdona: It would be good to find both solutions. I had considered maybe writing some shell functions/aliases and putting them in a separate repo as a kind of `fq` forensics toolkit. I like the first option you presented, displaying them expanded in place in the `sfl2` tree. As for how to display just the bookmark data, I'm not sure if it makes more sense to display them as separate JSON fragments, or as an array of objects.

dgmcdona:
I'm confused about why this won't work: 
```
fq 'grep_by(format="apple_bookmark")' test.plist
```
But this will:
```
fq 'grep_by(format=="jpeg")' some_file.bin
```
Does `fq` automatically probe raw bit fields when they are encountered? Is this how `format` works?

dgmcdona:
image
`fq -h apple_bookmarkdata` seems to be producing some strange effects with line endings. Is this something that is being caused by my editor? I noticed that the format markdown docs for other formats are much more succinct, should I remove all of the explanation of the format and maybe just leave them as references?

wader:
> `fq -h apple_bookmarkdata` seems to be producing some strange effects with line breaks. Is this something that is being caused by my editor? I noticed that the format markdown docs for other formats are much more succinct, should I remove all of the explanation of the format and maybe just leave them as references?

Hmm that is strange, i think your markdown should be ok. There is some line break code that is a bit shaky, i suspect that might be the reason (in markdown.jq), will investigate.

wader:
> You're right, UID was reading short. I fixed that so that it reads correctly now. I'm still seeing bit ranges unaccounted for, but a side-by-side comparison of the current `torepr` display and the output of `plutil -p` doesn't show any discrepancies or missing data.

👍  will have a look again also and see if i can figure something out, but i have run into formats that do have gaps that are not "reachable" or not even padding/alignment it seems

wader:
> I'm confused about why this won't work:
> 
> ```
> fq 'grep_by(format="apple_bookmark")' test.plist

`format` is a function in fq that takes no argument (so you leave out the `()`) and it returns name of the format if possible, an example probably makes it easier to explain:

```
$ fq format format/bplist/testdata/Info.plist
"bplist"
$ fq '.header | format' format/bplist/testdata/Info.plist
null
$ fq -n '123 | format'
null
```

So it only return strings for decode value "roots", this way it can be used to find all nested formats etc.

As a side note, there happens to exist a `format/1` function in jq that can be used to encode strings in different format (html, urlencode etc). Didn't even know before i named the fq `format/0`, is not documented. But jq allows function name overloading so they do not collide.

Back to the query:

So let's break down `grep_by(format="apple_bookmark")`:

`format="apple_bookmark"` will be a lambda expressed passed to `grep_by`, `grep_by` is a shorthand for `.. | select(<some expression here>)?` you can see the definition in grep.jq (also `..` is shorthand for `recurse`).

So `grep_by` will recurse and evaluate `format="apple_bookmark` for each value, and if it evaluates to something true-ish and not throw error it will output the value (is done with `select(...)?`), in summary it's a recursive filter that ignored errors. `format="apple_bookmark` is an assignment so it will eval each side and then assign,  LHS `format` will eval to a string or some format name, RHS just a literal string, so you will get `null="apple_bookmark"` which is just `"apple_bookmark"` which is true-ish. End result i think is that you will see all values separately except if `format` returns a string then i think it will throw an error that will be ignored.

> ```
> 
> But this will:
> 
> ```
> fq 'grep_by(format=="jpeg")' some_file.bin
> ```
> 
> Does `fq` automatically probe raw bit fields when they are encountered? Is this how `format` works?

`some_file.bin` will be probed and if successful grep_by will output all nested jpeg:s that can be found as `==` will compare what `format` outputs with `"jpeg"`.

Hope that helped. I'm thinking about writing some kind of jq guide, so good to practice explaining how jq works :)

So for sfl2 files i think we have to figure out some good way to find bookmarks blobs. I was thinking maybe it's poossible to reconstruct the NSKeyedArchiver object tree (assume it's not a cyclic graph?) and then look for boomarks and decode things?

wader:
Fixed the help word breaking a bit https://github.com/wader/fq/pull/497. Now looks ok i think, but maybe can move large links inside the text to references? they are treated liked a very wide word :)

format/bookmark/bookmark.md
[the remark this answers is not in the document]
wader: > It would be good to find both solutions. I had considered maybe writing some shell functions/aliases and putting them in a separate repo as a kind of `fq` forensics toolkit. I like the first option you presented, displaying them expanded in place in the `sfl2` tree. As for how to display just the bookmark data, I'm not sure if it makes more sense to display them as separate JSON fragments, or as an array of objects.

fq forensics toolkit sounds great! maybe a good place to test things and if stable/reusable/generic enough some things can move into fq? i've actually thought about doing something similar but for media files, i have a bunch of mp4 related snippets that don't really fit into fq.

I got a bit excited about decoding the NSKeysArchived and made some progress. Still very rough:

This is with the snake_case fix for apple_bookmark torepr.

Put this in ns_keyed_archiver.jq
```
def from_ns_keyed_archiver:
  (  . as {"$objects": $objs, "$top": {root: $root_uid}}
  | def _f($id):
      ( . #debug({$id})
      | $objs[$id]
      # | debug
      | if type == "string" then .
        elif type == "number" then .
        else
          (. as {"$class": $class}
          | if $class == 13 then # NSDictionary?
              ( . as {"NS.keys": $ns_keys, "NS.objects": $ns_objects}
              | [$ns_keys, $ns_objects]
              | transpose
              | map(
                  ( . as [$k, $o]
                  | {key: _f($k), value: _f($o)}
                  )
                )
              # | debug
              | from_entries
              )
            elif $class == 58 then #?
              ( . as {"NS.objects": $ns_objects}
              | $ns_objects
              | map(_f(.))
              )
            else "class-\($class)"
            end
          )
        end
      );
    _f($root_uid)
  );
```

Now this works, lots of options and things but nice to see it seems to work. Does it look correct?
```
$ go run . -o bits_format=string -L . 'include "ns_keyed_archiver"; torepr | from_ns_keyed_archiver | (.. | select(.Bookmark)?.Bookmark) |= (apple_bookmark | torepr)' '/Users/wader//Library/Application Support/com.apple.sharedfilelist/com.apple.LSSharedFileList.RecentApplications.sfl2'
{
  "items": [
    {
      "Bookmark": {
        "983055": true,
        "creation_options": 536870912,
        "display_name": "Firefox.app",
        "target_cnid_path": [
          77548117,
          1260261
        ],
        "target_creation_date": 532186965,
        "target_flags": "\u0012\u0002\u0000\u0000\u0000\u0000\u0000\u0000\u001f\u0002\u0000\u0000\u0000\u0000\u0000\u0000\u001a\u0002\u0000\u0000\u0000\u0000\u0000\u0000",
        "target_path": [
          "Applications",
          "Firefox.app"
        ],
        "volume_creation_date": 687789381,
        "volume_flags": "\ufffd\u0000\u0000\u0000\u0001\u0000\u0000\u0000\ufffd\u0013\u0000\u0000\u0001\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000",
        "volume_is_root": true,
        "volume_name": "Macintosh HD",
        "volume_path": "/",
        "volume_size": 250685575168,
        "volume_url": "file:///",
        "volume_uuid": "D63F4409-BCCD-48AE-896E-317E982AD80E"
      },
      "CustomItemProperties": {},
      "Name": "Firefox.app",
      "uuid": "8FD3DE25-8DCD-4B3B-915C-1CD6BE90AFED",
      "visibility": 0
    },
... more items here ...
}
```

format/applebookmark/apple_bookmark.md
```diff
@@ -0,0 +1,42 @@
...
+  [Sarah Edwards'](https://www.mac4n6.com/blog/2016/1/1/manual-analysis-of-nskeyedarchiver-formatted-plist-files-a-review-of-the-new-os-x-1011-recent-items)
+  excellent research on the subject.
+
+Locating `bookmarkData` objects in plist files, especially those of the
+`NSKeyedArchiver` type, is a place where `fq`'s recursive searching really
+shines. Here, we recursively descend through the decoded tree, probing for and 
+selecting any `bookmark` blobs, then converting them to readable JSON with
+`torepr`:
+```
+fq '.. | select(format=="bookmark") | .map(. | torepr)' com.apple.LSSharedFileList.RecentApplications.sfl2
```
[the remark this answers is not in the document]
dgmcdona: Just added some to the `bplist` testdata and tests 👍

dgmcdona:
You're explanation was very helpful! Unfortunately my question was malformed (I meant to have a second `=` in there there to check for equality, not do assignment). 

But after thinking some more about things, I think I understand why `grep_by(format=="apple_bookmark")` does not work in this instance: the `bplist` decoder does not actually invoke the `apple_bookmark` decoder, so it has no idea that a bookmark decode root is there; `format` does not make that determination at runtime. Am I thinking about that correctly?

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,321 @@
...
+
+	// all fields are little-endian with the exception of the Date datatype.
+	d.Endian = decode.LittleEndian
+
+	// decode bookmarkdata header, one at the top of each "file",
+	// although these may be nested inside of binary plists
+	d.FieldStruct("header", func(d *decode.D) {
+		d.FieldUTF8("magic", 4, d.AssertStr("book", "alis"))
+		d.FieldU32("total_size")
+		d.FieldU32("unknown")
```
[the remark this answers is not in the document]
dgmcdona: That might be a good idea, since a lot of formats that have been reverse engineered are going to have `unknown` fields

format/bookmark/bookmark.md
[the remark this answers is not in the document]
dgmcdona: This is amazing, going to make some people very happy 😄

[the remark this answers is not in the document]
dgmcdona: Only thing is `elif $class == 58 then #?` should be `elif $class == 56 then #?`, unless they have changed the format. Had to tweak it to make it work on my machine.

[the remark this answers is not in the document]
dgmcdona: Seems like those class numbers might cause problems, unfortunately there is some variability. I've had to make tweaks to make it work on different files.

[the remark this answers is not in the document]
dgmcdona: Seems like those class numbers might cause problems, unfortunately there is some variability. This solution seems to be more flexible: 
```
def from_ns_keyed_archiver:
  (  . as {"$objects": $objs, "$top": {root: $root_uid}}
  | def _f($id):
      ( . #| debug({$id})
      | $objs[$id]
      #| debug
      | if type == "string" then .
        elif type == "number" then .
        else
          (. as {"$class": $class}
          | debug
          | if ."NS.keys" != null and ."NS.objects" != null then
              ( . as {"NS.keys": $ns_keys, "NS.objects": $ns_objects}
              | [$ns_keys, $ns_objects]
              | transpose
              | map
                (
                  ( . as [$k, $o]
                  | {key: _f($k), value: _f($o)}
                  )
                )
              | from_entries
              )
            elif ."NS.objects" != null then
              ( . as {"NS.objects": $ns_objects}
              | $ns_objects
              | map(_f(.))
              )
            else "class-\($class)"
            end
          )
        end
      );
    _f($root_uid)
  );
```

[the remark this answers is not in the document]
dgmcdona:
```
fq -o bits_format=string -L . 'include "ns_keyed_archiver"; torepr | from_ns_keyed_archiver | (.. | select(.Bookmark)?.Bookmark) |= (apple_bookmark | torepr)'
```
Some variation of this would be great as an alias, maybe `sfl2_bookmarks`? I'm open to ideas about how you want to structure snippets. It might make sense to put them into a separate repository so that people that want to use them but are running `fq` via brew/go install/etc can clone the repo without bringing it all of the other stuff.

wader:
Looked at the bplist unknown fields a bit again. will answer the other things later, short lunch break now :), but think we're making good progress!

I observed this pattern, some array or object that is skipped or been lost somehow?
Screenshot 2022-11-30 at 13 46 52

wader:
> But after thinking some more about things, I think I understand why `grep_by(format=="apple_bookmark")` does not work in this instance: the `bplist` decoder package does not explicitly invoke the `apple_bookmark` decoder, so it has no idea that a bookmark decode root is there; `format` does not make that determination at runtime. Am I thinking about that correctly?

Correct! and i guess it's tricky to make the bplist decoder somehow be able to detect "subformats" like bootmarks, even tricker when its yet another encoding like NSKeyedArchive. So doing it with some jq help function or some snippet if short enough make sense i think.

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,311 @@
...
+
+//go:embed apple_bookmark.jq apple_bookmark.md
+var bookmarkFS embed.FS
+
+func init() {
+	interp.RegisterFormat(decode.Format{
+		Name:        format.APPLE_BOOKMARK,
+		ProbeOrder:  format.ProbeOrderBinUnique,
+		Description: "Apple BookmarkData",
+		Groups:      []string{format.PROBE},
```
wader, line 21: Does it make sense to make the format probable? are the places were it's used in separate files? thinking if we will have some jq code that explicitly decodes it's not needed.

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,311 @@
...
+		d.FieldRawLen("reserved", reservedSize*8)
+	})
+
+	tocOffset := calcOffset(d.FieldU32("first_toc_offset"))
+
+	var currentHdr *tocHeader
+	var tocHeaders []*tocHeader
+
+	d.FieldArrayLoop("toc_headers", func() bool {
+		return tocOffset != headerEndBitPos
```
wader, line 292: Some limit on this loop? btw you can do `make fuzz GROUP=apple_bookmark` to fuzz just one format, have been a very efficient way to find infinite loop and excessive memory use

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,311 @@
...
+			tocOffset = calcOffset(currentHdr.nextTOCOffset)
+			tocHeaders = append(tocHeaders, currentHdr)
+		})
+	})
+
+	// now that we've collected all toc headers, iterate through each one's
+	// entries and decode associated records.
+	d.FieldArray("bookmark_entries",
+		func(d *decode.D) {
+
```
wader, line 305: Spurious blank line

format/applebookmark/testdata/apple_bookmark.fqtest
```diff
@@ -0,0 +1,960 @@
+$ fq dv sample1.book
```
wader, line 1: I've started to have the test files as `<FILE>` and `<FILE>.fqtest` pairs, think it makes a bit cleaner. Maybe good here?

wader:
Looks good overall i think, some things you want to add? the jq snippet code things?

format/applebookmark/apple_bookmark.go
dgmcdona: Nice, `make fuzz` caught a case of infinite recursion in the `decodeRecord` function which makes sense. Added a check for that and set what I think is a reasonable limit of 100 TOCs here as well.

format/applebookmark/apple_bookmark.go
dgmcdona: I haven't seen them outside of the `bplist` test files I've been working with, but they can be created as part of Apple's public APIs so it's possible that they might be? Some quick googling didn't tell me much about how other people are storing these.

format/bookmark/bookmark.md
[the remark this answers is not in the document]
dgmcdona: Do we want to put `ns_keyed_archiver.jq` in the `bplist` package? And if so, do we need to register the function in the same way that we do `torepr`?

dgmcdona:
Found definitions for the individual bit fields so I added decoding for that. Those are tricky but I *think* they are being decoded correctly. It's difficult to test and verify.

format/applebookmark/apple_bookmark.go
wader: Okok, i'm mostly concerned about have decoders in the probe group if they can mis-probe or for some reason will take long time to fail slowing down probing a lot

wader:
A bit unrelated to this PR but what are some similar tools to fq for doing forensics? think i would be interesting to see how they work

format/bookmark/bookmark.md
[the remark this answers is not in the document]
wader: I think we can include ns_keyed_archiver.jq for bplist, it seem to be very common usage for plists? but it feels a bit rough atm, maybe we can work on it separate to this PR? for example i wonder how these class numbers work, are they stable? maybe the ones from cocoa at least? if not i guess we have to have an argument etc that is used to select mapping?

There is no need to register functions, the will just appear. But `torepr` is a bit special as it is overloaded per format, if your curious it's implemented here https://github.com/wader/fq/blob/012e7527f05508e1d154a9cf18de4e9b8b216686/pkg/interp/funcs.jq#L159 with help of https://github.com/wader/fq/blob/master/pkg/interp/format_func.jq

dgmcdona:
https://github.com/ydkhatri/mac_apt
https://ericzimmerman.github.io/#!index.md
https://github.com/simsong/bulk_extractor

These are a few. Some have a different mission than `fq` in that they scan entire live systems for artifacts and collect them, but fundamentally they still have to parse the binary formats at some point. I think this PR actually improves upon the bookmarkData parsing in `mac_apt`. I have been working on `lnk` files in a separate branch, and I'm thinking Windows Prefetch files would be the next goal after that. That might be tricky due to compression but I know `fq` does gzip so I'm sure there's a way to handle it. 

The fact that `fq` can be used cross-platform as a standalone executable sets it apart from many of the tools that use Python, which requires managing dependencies/environments, or the ones that require .NET. Plus, it's fast. I'm not a forensics practitioner so I don't know how much ground needs to be covered for this to become a go-to part of the standard toolset, but I want to just keep adding things and see if people want to use it.

format/applebookmark/apple_bookmark.md
```diff
@@ -0,0 +1,41 @@
+## Apple bookmarkData format
```
wader, line 1: You can skip the header, it's added by make doc and the help system. It will use the description.

I usually have a `help_<format>.fqtest` file in testdata with just `$ fq -h <format>` to test it.

format/applebookmark/apple_bookmark.md
```diff
@@ -0,0 +1,41 @@
...
+  For more information about these types of files, see Sarah Edwards' excellent
+  research on the subject (link in references).
+
+Locating `bookmarkData` objects in plist files, especially those of the
+`NSKeyedArchiver` type, is a place where `fq`'s recursive searching really
+shines. Here, we recursively descend through the decoded tree, probing for and 
+selecting any `bookmark` blobs, then converting them to readable JSON with
+`torepr`:
+```
+fq '.. | select(format=="bookmark") | .map(. | torepr)' com.apple.LSSharedFileList.RecentApplications.sfl2
```
wader, line 29: Guess we have to fix this snippet somehow

wader:
> https://github.com/ydkhatri/mac_apt https://ericzimmerman.github.io/#!index.md https://github.com/simsong/bulk_extractor https://github.com/sleuthkit/sleuthkit
> 
> These are a few. Some have a different mission than `fq` in that they scan entire live systems for artifacts and collect them, but fundamentally they still have to parse the binary formats at some point. I think this PR actually improves upon the bookmarkData parsing in `mac_apt`.

Thanks, had quick look at some of them, interesting stuff, will look more later.

Currently my own focus for fq has been on very detailed decoding, presentation of the data and querying. Making thing possible to do over raw performance. I use it as an assistant to quickly show me things, test hypothesis and such things when developing and debugging software that reads/writes various binary formats.

> I have been working on `lnk` files in a separate branch, and I'm thinking Windows Prefetch files would be the next goal after that. That might be tricky due to compression but I know `fq` does gzip so I'm sure there's a way to handle it.

Sounds good!

Do you know what kind of compression? as long as it's available as go code it shouldn't be a problem. As you noticed fq has support for sub/nested binary buffers in the same decode tree which i think is quite neat.

> The fact that `fq` can be used cross-platform as a standalone executable sets it apart from many of the tools that use Python, which requires managing dependencies/environments, or the ones that require .NET. Plus, it's fast. I'm not a forensics practitioner so I don't know how much ground needs to be covered for this to become a go-to part of the standard toolset, but I want to just keep adding things and see if people want to use it.

Most of the cross-platform support i have to thank golang for :) it also helps that fq currently do not need much from the platforms it runs on, read files and some tty stuff.

Keep adding! happy to help out, and i'm very grateful that you want to contribute. Also very helpful to have someone poke at your stuff, so don't hesitate question how things work or ask things, there is plenty of unfinished and hacky things in the projects.

wader:
btw it would be helpful to prefix commit messages with one or more contexts, makes it easier to search and do release notes. I usually use format name, "doc", "help", "decode" etc.

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+const (
+	arrayEntrySize = 4
+	dictEntrySize  = 4
+)
+
+type stack []int64
+
+var offsetStack stack
```
wader, line 328: Neat solution, like it. Can see this being used in more places, might move it into decode.go if so 👍

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+	*s = append(*s, i)
+}
+
+func (s *stack) pop() {
+	*s = (*s)[:len(*s)-1]
+}
+
+func decodeRecord(d *decode.D) {
+	offsetStack.tryPush(d)
```
wader, line 345: A quite nice API design is push returns a function that pop, then on can do `defer stack.PushAndPop(d)()`

format/applebookmark/apple_bookmark.jq
```diff
@@ -0,0 +1,41 @@
...
+      elif .type == "double" then .data | tovalue
+      elif .type == "date" then .data | tovalue
+      elif .type == "boolean_false" then false
+      elif .type == "boolean_true" then true
+      elif .type == "flag_data" then
+        ( .enabled_property_flags as $eflags 
+          | .property_flags 
+          | to_entries
+          | map(select($eflags[.key] and (.key | startswith("reserved") | not)))
+          | from_entries
```
wader, line 19: Indent left one? I see you get a hang of jq! nice :) btw there is a `with_entries` to do `to_entries | f | from_entries` kindof but i usually like it explicit like this

wader: Also if you want jq IDE support and documentation, i use https://github.com/wader/jq-lsp, https://github.com/wader/vscode-jq and https://github.com/wader/jq-dash-docset more it a lot nicer to work with jq

dgmcdona: Thanks, wish I had set all of that up sooner!

wader: 👍  sorry i should have mention it earlier 😄 the syntax check and also check for undefined functions and variables is very nice to have. Note that when working with fq related jq-code it might show undefined function errors at times as the jq code in fq runs in a bit different environment (can be workaround with a `.jq-lsp.jq` file but maybe not worth it).

The install of it is a bit cumbersome, hope it was ok? I should really try to make it nicer, fix binaries for jq-lsp and add vscode-jq to the vscode package repository thingy.

dgmcdona: It was pretty easy. I use neovim with https://github.com/neoclide/coc.nvim and just added this to my config, it works perfectly:
```
{
  "languageserver": {
    "jq": {
      "command": "jq-lsp",
      "filetypes": ["jq"]
    }
  }
}
```
```

File: format/bookmark/bookmark.md (hunk 1 above)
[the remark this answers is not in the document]
Comment (dgmcdona) [reply format/bookmark/bookmark.md]: Ok, I'll set it aside for the moment and open an issue for it.

File: format/applebookmark/apple_bookmark.jq (hunk 15 above)
Comment (wader) [reply format/applebookmark/apple_bookmark.jq:19]: Wow did not expect it to work with neovim just like that :) now i have to try it

File: format/applebookmark/apple_bookmark.go [hunk 16]
```diff
@@ -0,0 +1,455 @@
...
+
+var elementTypeMap = scalar.UToScalar{
+	elementTypeTargetURL:             {Sym: "target_url", Description: "A URL"},
+	elementTypeTargetPath:            {Sym: "target_path", Description: "Array of individual path components"},
+	elementTypeTargetCNIDPath:        {Sym: "target_cnid_path", Description: "Array of CNIDs"},
+	elementTypeTargetFlags:           {Sym: "target_flags", Description: "flag bitfield"},
+	elementTypeTargetFilename:        {Sym: "target_filename", Description: "String"},
+	elementTypeCNID:                  {Sym: "target_cnid", Description: "4-byte integer"},
+	elementTypeTargetCreationDate:    {Sym: "target_creation_date", Description: "Date"},
+	elementTypeUnknown1:              {Sym: "unknown", Description: "Unknown"},
```
wader, line 114: Maybe sym should be unknown1, unknown2, ...?

format/applebookmark/apple_bookmark.go
wader: Let's keep in probe for now, as it assert early on book/alia it feels kind of safe

wader:
Had some tiny comment and maybe add the help_<format>.fqtest. After that i think we can merge. Something you want to add?

wader:
I've noticed that formats.svg causes conflicts a lot so now part of #503 you can do just `make README.md doc/formats.md` to regenerate and just skip formats.svg, i will update it from time to time.

wader:
Think after you regenerate formats.md and this is merged i will add some padding to the format nam column so that conflicts gets less likely there also

format/applebookmark/apple_bookmark.jq
```diff
@@ -0,0 +1,39 @@
...
+        ( .data
+        | map({key: (.key | _f), value: (.value | _f)})
+        | from_entries
+        )
+      elif .type == "uuid" then .data | tovalue
+      elif .type == "url" then .data | tovalue
+      elif .type == "relative_url" then
+        .data | map(.record.data)
+      end
+    );
```
wader, line 35: If you want i think you can remove the `( ... )` around the `if` here

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,455 @@
...
+func decodeRecord(d *decode.D) {
+	defer offsetStack.pushAndPop(d)()
+
+	d.FieldStruct("record", func(d *decode.D) {
+		n := int(d.FieldU32("length"))
+		typ := d.FieldU32("type", dataTypeMap)
+		switch typ {
+		case dataTypeString:
+			d.FieldUTF8("data", n)
+			d.FieldRawLen("alignment_bytes", 32-((d.Pos()+32)%32))
```
wader, line 354: Could possibly use `d.AlignBits(32)`

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,455 @@
...
+
+		})
+	}
+}
+
+func decodeTOCHeader(d *decode.D) *tocHeader {
+	hdr := new(tocHeader)
+
+	d.FieldStruct("toc_header", func(d *decode.D) {
+		hdr.tocSize = d.FieldU32("toc_size")
```
wader, line 305: tocSize not used? numEntries enough?

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,455 @@
...
+		})
+		hdrCount++
+	})
+
+	// now that we've collected all toc headers, iterate through each one's
+	// entries and decode associated records.
+	d.FieldArray("bookmark_entries",
+		func(d *decode.D) {
+			for _, hdr := range tocHeaders {
+				d.SeekAbs(hdr.entryArrayOffset, hdr.decodeEntries)
```
wader, line 451: Was confused how number of entries was pass to decodeEntries, didn't notice decodeEntries was a method, nice :)

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,455 @@
...
+
+func bookmarkDecode(d *decode.D, _ any) any {
+	// all fields are little-endian with the exception of the Date datatype.
+	d.Endian = decode.LittleEndian
+
+	// decode bookmarkdata header, one at the top of each "file",
+	// although these may be nested inside of binary plists
+	d.FieldStruct("header", func(d *decode.D) {
+		d.FieldUTF8("magic", 4, d.AssertStr("book", "alis"))
+		d.FieldU32("total_size")
```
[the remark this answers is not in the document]
dgmcdona: I think it's okay as it is, I removed it `tocSize` from the struct. Whats the intended use case for `FramedFn`?

format/applebookmark/apple_bookmark.go
dgmcdona: I tried that but it was resulting in gaps since fields weren't being produced. Is something like this the way to do it?
```
d.FieldSFn("padding", func(d *decode.D) int64 { return int64(d.AlignBits(32)) })
```

dgmcdona:
Any suggestions for a good workflow for when I need to regenerate `fqtest` output after decoder changes? It's a bit tedious to go through, delete the old stuff, and rerun them individually; I suspect you might have a better way to do that.

format/applebookmark/apple_bookmark.md
dgmcdona: Having some issues with this related to the long lines again. `make test` is producing output where the lines have been broken to a fixed width, whereas the output I put into the `fqtest` file has the long lines unbroken.

format/applebookmark/apple_bookmark.go
wader: Oh sorry now i see, this pads to make the record size an even 4 byte size? `d.FieldRawLen("alignment_bytes", int64(d.AlignBits(32)))` would pad to align to a 32 bit position

format/applebookmark/apple_bookmark.go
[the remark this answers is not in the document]
wader: 👍 

`FramedFn(nBits, f)` decodes with a limit how far `f` can read and once done seeks to end of frame. There is also `LimitedFn(nBits, f)` that do the same but don't touch the position. "frame" name is inspired from media containers where samples get their "framing" from the container but is not a great name, still looking for better names or API design for this, suggestions welcome.

doc/formats.md
```diff
@@ -2,126 +2,127 @@
 
 [fq -rn -L . 'include "formats"; formats_table']: sh-start
```
wader, line 4: Now i see why it still caused conflict, apple_bookmark is the new longest format name :)

format/applebookmark/apple_bookmark.md
wader: > Any suggestions for a good workflow for when I need to regenerate `fqtest` output after decoder changes? It's a bit tedious to go through, delete the old stuff, and rerun them individually; I suspect you might have a better way to do that.

Oh you copy/paste it manually, that should not be needed. I usually run something like this to update the expected output:
```
WRITE_ACTUAL=1 go test -v -run TestFormats/applebookmark ./format
```
Then it will rewrite the .fqtest files with the current output, you can verify after by running without `WRITE_ACTUAL`.

So my workflow for tests is usually that is create a fqtest file with just `$ fq ...` lines and then run the tests with `WRITE_ACTUAL` to get the output.

Sometimes when i'm quite sure a change needs an update is probably correct i run directly with `WRITE_ACTUAL` and verify with `git diff` before `git add`.

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,454 @@
...
+}
+
+const (
+	arrayEntrySize = 4
+	dictEntrySize  = 4
+)
+
+type stack []int64
+
+var offsetStack stack
```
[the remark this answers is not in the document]
dgmcdona: Yeah, I don't know what I was thinking making that global. Also a problem in `pushAndPop`, the method references the global variable instead of the struct pointer.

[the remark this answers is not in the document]
wader: > Yeah, I don't know what I was thinking making that global. Also a problem in pushAndPop, the method references the global variable instead of the struct pointer.

No worries. Im surprised it was not detect by the test running with -race hmm

>After some thought the cleanest way to do things was to make the stack internal to the decode.D. Maybe this is good going forward? If so, can fix up the bplist decoder to use it as well since it does something similar.

I'm a bit reluctant to add things to `*decode.D`, need to think a bit more about it if so. I guess that field would have to get inherited down into sub decoders? could we go with a separate type for this for now? maybe we could put it in the decode package if it should be shared? another option is to put bplist and apple_bookmark into formart/apple etc so they can share things?

What do you think about something like this:

```go
```

wader:
Sorry for all the late comments, hard to review by just reading so once you start poking around in an editor you notice things :)

But think we're close to merge now

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,435 @@
...
+
+			// if the key has the top bit set, then (key & 0x7fffffff)
+			// gives the offset of a string record.
+			if entry.key&0x80000000 != 0 {
+				d.FieldStruct("key_string", func(d *decode.D) {
+					d.SeekAbs(calcOffset(entry.key&0x7fffffff), decodeRecord)
+				})
+			}
+
+			entry.recordOffset = calcOffset(d.FieldU32("offset_to_record"))
```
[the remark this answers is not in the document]
dgmcdona: I think we have to leave this as-is so that `torepr` will continue to work properly (it relies on the key translations in the `elementTypeMap`).

format/applebookmark/apple_bookmark.go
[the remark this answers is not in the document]
dgmcdona: For some reason it is not catching the infinite recursion in the example that I produced with your `fq` one-liner there. See the `loop.book` that I added to the testdata.

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,435 @@
...
+type dataObjectDecoder struct {
+	flagFn func(d *decode.D)
+}
+
+func (dod *dataObjectDecoder) decode(d *decode.D) {
+	decodeFlagDataObject(d, dod.flagFn)
+}
+
+var resourcePropDecoder = &dataObjectDecoder{decodeTgtPropertyFlagBits}
+var volumePropDecoder = &dataObjectDecoder{decodeVolPropertyFlagBits}
```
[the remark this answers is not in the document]
wader: 👍 I guess we could try use generics to make this pattern nicer. It seems to happen quite a lot, maybe something like `d.SeekAbsFn1<T>(offset, some-T-value, fn(d *decode.D, v T) { ... })` could save on nested functions in the decoder code (and have SeekAbsFn2 if it takes two generic args)

format/applebookmark/apple_bookmark.go
```diff
@@ -0,0 +1,448 @@
...
+
+func makeDecodeRecord() func(d *decode.D) {
+	var pld posLoopDetector
+
+	var decodeRecord func(d *decode.D)
+	decodeRecord = func(d *decode.D) {
+		defer pld.pushAndPop(
+			d.Pos(),
+			func() { d.Fatalf("infinite recursion detected in record decode function") },
+		)
```
wader, line 337: typo here, pop function is not defer called, add a `()`, after that the loop fqtest works for me 👍

format/applebookmark/apple_bookmark.go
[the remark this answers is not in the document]
wader: > For some reason it is not catching the infinite recursion in the example that I produced with your fq one-liner there. See the loop.book that I added to the testdata.

Was a typo, see code comment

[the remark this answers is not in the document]
wader: > This is really cool, I was trying to figure out the cleanest way to add the stack to the scope of the decodeRecord function that didn't make a huge mess when passing parameters around, but I didn't know closures could be used like this!

Yeap very useful pattern! go has quite good support for closures and lambdas, being GC:d and blurring what is stack and heap helps also. The syntax is a bit verbose, but not sure i wish for ml/haskell syntax either, a tiny bit more syntax help would be nice :)

wader:
> Any suggestions for a good workflow for when I need to regenerate fqtest output after decoder changes? It's a bit tedious to go through, delete the old stuff, and rerun them individually; I suspect you might have a better way to do that.

Did you get the WRITE_ACTUAL workflow to work?

Some other workflows i use:

Use `watchexec` etc so i don't have to switch terminals so much, ex run fq with some nice query in a separate terminal:
```
# -c clear screen
watchexec 'go run . torepr format/applebookmark/testdata/sample1.book'
watchexec -c 'go run . torepr format/applebookmark/testdata/sample1.book'
watchexec -c -- go run . torepr format/applebookmark/testdata/sample1.book
```

Use `LOGFILE`:
```
LOGFILE=/tmp/fq go run . -d blabla ...
```
Now `log.Printf` and friends output will be written to `/tmp/fq`, and then i usually run a separate terminal with `tail -f /tmp/fq`. This is very useful when the output from fq gets to mixed up.

It's usually also nice in general to spend some time to get a query that shows just the part your work on and then start to do changes.

wader:
Ready to merge? only nit pick left was to possibly replace `tocEntry` with local vars in `decodeEntries`.

dgmcdona:
One last question though - I keep getting test failures locally for three formats, all for the same reason (differences in the way raw byte ranges are represented as strings). Something misconfigured locally maybe?
```
    --- FAIL: TestFormats/mp4/testdata/in24.fqtest (0.10s)
        fqtest.go:15:
            --- expected
            +++ actual
            @@ -111,7 +111,7 @@
             0x240|00 00 00 00                                    |....            |                  modification_time: 0 (1904-01-04T00:00:00Z) 0x240-0x243.7 (4)
             0x240|            00 00 ac 44                        |    ...D        |                  time_scale: 44100 0x244-0x247.7 (4)
             0x240|                        00 00 00 2c            |        ...,    |                  duration: 44 0x248-0x24b.7 (4)
            -0x240|                                    7f ff      |            ..  |                  language: "\x7f\x7f\x7f" 0x24c-0x24d.7 (2)
            +0x240|                                    7f ff      |            ..  |                  language: "\u007f\u007f\u007f" 0x24c-0x24d.7 (2)
             0x240|                                          00 00|              ..|                  quality: 0 0x24e-0x24f.7 (2)
                  |                                               |                |                [1]{}: box 0x250-0x27c.7 (45)
             0x250|00 00 00 2d                                    |...-            |                  size: 45 0x250-0x253.7 (4)
```

wader:
🥳

dgmcdona:
> 

Upgraded to 1.19 and that fixed it, thx

wader:
Great, but maybe a workaround is worth it, official go 1.18 support will be dropped in 4 month i think but i suspect that new versions of fq will be able to build with it for a long time.

PR #499, opened 2022-12-01
interp: Change default bits_format=string

wader:
I think this is more intuitive but might in some case cause very large JSON output but maybe that less common or expected. In does cases i think you either want to use some other bits_format (md5, truncate, etc) or you delete/transform the jq value before turn it into JSON.

Strings in gojq are binary safe so you can use to hold raw bytes. But note that convert the binary into JSON is lossy, same as the JSON standard.

Add bits_format option documentation.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -699,6 +699,22 @@ Text encodings
 - `toutf16be` Encode string as UTF16 big-endian into binary.
 - `fromutf16be` Decode binary as UTF16 big-endian into string.
 
+## Options
+
+fq has some general options in addition to decode and decoders specific options. They all use the same `-o <name>=<value>` argument.
+
+`<value>` is fuzzily parsed based on the type of the option. Ex: a string can specified as `-o name=string` or `-o name="string"`.
+
+### `bits_format`
+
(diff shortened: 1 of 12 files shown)


wader:
@dgmcdona This should make it easier to further transform decode fields like for bplist/apple_bookmark

dgmcdona:
What do you think about some sort of `.fq.conf` file so users can set options to match their own common cases without having to specify each time?

wader:
> What do you think about some sort of `.fq.conf` file so users can set options to match their own common cases without having to specify each time?

Yeah that is good idea, have a couple of way of the top of my head:
- Something in the already existing `init.jq` file, but it would have to be some kind of jq syntax, maybe not easy to use.
- Environment variable with options `FQ_OPTIONS` etc with raw `-o ...` args? or `name=value,name=value...` or `name=value name=value `
- A separate file like `fq.conf` somehow
- More?

In either case i think command line argument should override, so order would defaults,user,cli-args,repl-options,function-options (to `d` etc)?

btw if you want to see all options with current value you can do `fq -n options`, some of them are a bit internal.

PR #500, opened 2022-12-01
decode,interp: Rename unknown gap fields from "unknown#" to "gap#"

wader:
Think it makes it clearer and also less likely to collide with a field name a deocder wants to use.


wader:
@dgmcdona was good idea, was reluctant at first now i like it :)

PR #503, opened 2022-12-04
dev,doc,make: Cleanup makefile and have proper targets for *.md and *.svg

wader:
Makes easier to not cause formats.svg conflicts


wader:
@bitbears-dev this should fix conflicts issues less likely, switch to `make md-doc` and i will do `make doc` from time to time

wader:
@bitbears-dev change of plans, i added proper make target instead so now `make README.md doc/formats.md` works. Problem with formats.svg is that different graphviz versions seems produce slightly different outputs :(

PR #516, opened 2022-12-10
interp,display: Add workaround for go 1.18 when escaping 0x7f

wader:


wader:
@dgmcdona hacky but probably worth it

PR #518, opened 2022-12-11
decode: fixes endian of unicode strings

dgmcdona:
`bplist`: Quick fix that corrects the endianness of Unicode strings.


wader:
👍

PR #523, opened 2022-12-14
decode: More type safe API and split scalar into multiple types

wader:
Preparation to make decoder use less memory and API more type safe. Now each scalar type has it's own struct type so it can store different things and enables to have a scalar interface.
Also own types will enable experimenting with decode DLS designs like using chained methods that are type aware.


wader:
@dgmcdona this might break some unmerged code but should be quite straight to get working again, prefix some methods with scalar type, t.ex `scalar.Mapper` is now `scalar.UintMapper`, `scalar.StrMapper`, `d.Assert` is now `d.UintAssert`, `scalar.UToScalar` is now `scalar.UintMap`. `scalar.UToSymStr` is now `scalar.UintMapSymStr` etc.

wader:
@pnsafonov probably relevant for you also, also relevant as a step towards making fq more memory efficient. this PR only makes it a bit more efficient but now it should be easier to try things

Issue #502, opened 2022-12-03
bplist: NSKeyedArchiver jq function


dgmcdona:
`NSKeyedArchiver` stores objects in a `bplist` format by flattening the object into a set of keys and values, which reference each other by index. A common example of these are the `sfl2` files located in `~/Library/Application Support/com.apple.sharedfilelist`. @wader proposed the following function for reconstructing these objects into a more meaningful JSON representation: 
```jq
def from_ns_keyed_archiver:
  (  . as {"$objects": $objs, "$top": {root: $root_uid}}
  | def _f($id):
      ( . #debug({$id})
      | $objs[$id]
      # | debug
      | if type == "string" then .
        elif type == "number" then .
        else
          (. as {"$class": $class}
          | if $class == 13 then # NSDictionary?
              ( . as {"NS.keys": $ns_keys, "NS.objects": $ns_objects}
              | [$ns_keys, $ns_objects]
              | transpose
              | map(
                  ( . as [$k, $o]
                  | {key: _f($k), value: _f($o)}
                  )
                )
              # | debug
              | from_entries
              )
            elif $class == 58 then #?
              ( . as {"NS.objects": $ns_objects}
              | $ns_objects
              | map(_f(.))
              )
            else "class-\($class)"
            end
          )
        end
      );
    _f($root_uid)
  );
```

However, it was found that the class numbers are not consistent across multiple files, so relying on them for interpreting underlying types is not a general solution. The following seems to work:

```jq
def from_ns_keyed_archiver:
  (  . as {"$objects": $objs, "$top": {root: $root_uid}}
  | def _f($id):
      ( . #| debug({$id})
      | $objs[$id]
      #| debug
      | if type == "string" then .
        elif type == "number" then .
        else
          (. as {"$class": $class}
          | . #debug
          | if ."NS.keys" != null and ."NS.objects" != null then
              ( . as {"NS.keys": $ns_keys, "NS.objects": $ns_objects}
              | [$ns_keys, $ns_objects]
              | transpose
              | map
                (
                  ( . as [$k, $o]
                  | {key: _f($k), value: _f($o)}
                  )
                )
              | from_entries
              )
            elif ."NS.objects" != null then
              ( . as {"NS.objects": $ns_objects}
              | $ns_objects
              | map(_f(.))
              )
            else "class-\($class)"
            end
          )
        end
      );
    _f($root_uid)
  );
```
[...]

PR #527, opened 2022-12-17
decode: add ns_keyed_archiver, restructure apple decoders into package

dgmcdona:
Makes a tweak to the `bplist` `torepr` function to preserve type information when representing `cfuid` types. This enables the new `from_ns_keyed_archiver` function to correctly distinguish object references from integers.

Additionally, it restructures the apple decoders (`bplist`, `bookmark`, and `macho`) into a single directory so that some code may be shared.


dgmcdona:
Does jq allow for functions to have an optional parameter? In this case, I'd want `from_ns_keyed_archiver` to use `$top.root` if no parameter is supplied, or decode from a given index if the user passes it as a parameter.

format/apple/bookmark/apple_bookmark.go
```diff
@@ -299,32 +300,12 @@ const (
 	dictEntrySize  = 4
 )
 
-type posLoopDetector []int64
```
dgmcdona, line 302: I was hoping to use generics for this but then discovered that Go doesn't support generics in method signatures ☹️  I don't know if there is a clever trick that I'm missing that could make this work well, but it would be nice if this could be generic for all type satisfying `constraints.Integer`.

wader: Yes it's a bit annoying limitation, but seems like a tricky thing to mix with interfaces :( is also the reason why i haven't used generics much for the decode code yet. Think i will stick with generating code instead for now, which also makes it very flexible... but becomes a bit complex.

In the case of `posLoopDetector` could the generics type be in on the type? something like:
```
type posLoopDetector[T constraints.Integer] []T

func (t *posLoopDetector[T]) a() {
	...
}
```
Or you mean something else?

wader:
> Does jq allow for functions to have an optional parameter? In this case, I'd want `from_ns_keyed_archiver` to use `$top.root` if no parameter is supplied, or decode from a given index if the user passes it as a parameter.

Yes you can do it via different arity, ex:
```jq
def f(arg): ...
def f: f(<default value>);
```
Not that there is subtle difference between `f(a)` and `f($a)`, without `$` it's a lambda, with `$` is a binding, it usually does not make that much difference unless you use the arg multiple times in the function, and for a bindings the function will be called for each output of the argument expression: ex for `f(1,2,3)` if arg is a binding it will be called three times with the binding being 1 then 2 then 3, as a lambda it will be called once but once you use the arg inside the function it will output three values.

```
$ jq -cn 'def f(a): [a]; f(1,2,3)'
[1,2,3]
$ jq -cn 'def f($a): [$a]; f(1,2,3)'
[1]
[2]
[3]
```

format/apple/bplist/ns_keyed_archiver.jq
```diff
@@ -0,0 +1,70 @@
+def from_ns_keyed_archiver:
+  (  . as {
```
wader, line 2: Tiny left indent

format/apple/bplist/ns_keyed_archiver.jq
```diff
@@ -0,0 +1,70 @@
...
+      "$objects": $objects,
+      "$top": {root: $root}
+      #"$top": {"796BFF22-6712-4486-A32C-A1C5DB3273BA": $root}
```
wader, line 5: Hmm guess need to clean this up somehow

format/apple/bookmark/apple_bookmark.go
dgmcdona: Thanks, that's pretty much what I was trying to do but I was putting the generic type in the wrong place.

format/apple/bplist/ns_keyed_archiver.jq
dgmcdona: Overloaded `from_ns_keyed_archiver` so that the user can specify an optional root value from which to begin decoding

format/apple/bplist/ns_keyed_archiver.jq
dgmcdona: I think I fixed it, hopefully at some point can figure out some kind of formatting solution for neovim.

format/apple/bplist/ns_keyed_archiver.jq
```diff
@@ -0,0 +1,77 @@
...
+    def _f($id): _f($id; {"\($id)": true});
+    _f(root)
+  );
+
+def from_ns_keyed_archiver:
+  ( . as 
+      {
+        "$top": {root: $root}
+      }
+    | from_ns_keyed_archiver($root.cfuid)
```
wader, line 76: Could possibly be just i think?
```jq
def from_ns_keyed_archiver:
  from_ns_keyed_archiver(."$top".root.cfuid);
```

format/apple/bplist/ns_keyed_archiver.jq
```diff
@@ -0,0 +1,77 @@
+def from_ns_keyed_archiver(root):
```
wader, line 1: Maybe good to make this `$root` as we use it as a value?

format/apple/bplist/ns_keyed_archiver.jq
dgmcdona, line 76: Maybe this should be `from_ns_keyed_archiver($root?.cfuid // 1)` as a sensible fallback option? Or is it better if it just fails?

format/apple/bplist/ns_keyed_archiver.jq
```diff
@@ -0,0 +1,77 @@
...
+  (  
+    . as 
+      {
+        "$objects": $objects
+      }
```
wader, line 6: "fq-jq-style" would be this i think:
```jq
  ( . as {
      "$objects": $objects
    }
  | ...
```
But maybe we can skip fixing style stuff and i can re-format things later. Have thought about doing a jqfmt thingy but it's quite tricky :)

wader (commented):
Add `from_ns_keyed_archiver` usage examples to doc?

format/apple/bplist/ns_keyed_archiver.jq
wader: Aha that might explain things, vscode-jq does a quite ok job based on the highlighting syntax it seems... but is not great.

format/apple/bplist/ns_keyed_archiver.jq
wader: Are there files in the wild that seems assume root uid 1? could it be misleading if not and it works?

wader: Ah now i see. Now you have to write something like:
```
fq 'torepr | from_ns_keyed_archiver(."$top"."796BFF22-6712-4486-A32C-A1C5DB3273BA"."cfuid")'  ~/Downloads/com.apple.networkextension.plist
```

dgmcdona: `com.apple.networkextension.plist` is a good example of one that doesn't have a `$root` value and begins decoding at index 1, but I do lean towards forcing the user to specify an alternative root so that there aren't potential false positives and missing data in the resulting representation.

wader: And lookin for a UUID key might quite hacky? maybe best to default to 1 then?

wader: Ok, let's not default then 1 then. Maybe then should throw some more helpful error on null $root? and possibly add example to doc about uid 1?

dgmcdona: Much better, thanks! Also added an error case here.

dgmcdona:
I think this is ready to go, I've checked all of the types listed [here](https://github.com/strozfriedberg/plistutils/blob/fc7783449da1ed222547ceb5c416402216fa9b34/plistutils/nskeyedarchiver.py#L130) and I think that the `else` case handles the general dictionary types, so there's no need to implement special decoding for things like `SFLListItem`.

format/apple/bplist/bplist.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+### Decoding NSKeyedArchiver serialized objects
+
+A common way that Swift and Objective-C libraries on macOS serialize objects
+is through the NSKeyedArchiver API, which flattens objects into a list of elements
+and class descriptions that are reconstructed into an object graph using CFUID
+elements in the property list. `fq` includes a function, `from_ns_keyed_archiver`,
+which will rebuild this object graph into a friendly representation. The `bplist`
+data must first be passed through `torepr` to create a JSON representation.
```
wader, line 37: One could in `from_ns_keyed_archiver` check if the input is a "bplist" decode value and do `torepr` autoamatically (something like `if _exttype == "decode_value" and format == "bplist" then torepr end ...`) but maybe the function is already getting complicated. Maybe possibly for later?

format/apple/bplist/ns_keyed_archiver.jq
```diff
@@ -0,0 +1,71 @@
...
+            end
+          )
+        end
+      );
+    def _f($id): _f($id; {"\($id)": true});
+    _f($root)
+  );
+
+def from_ns_keyed_archiver:
+    from_ns_keyed_archiver(."$top"?.root?.cfuid // error("root node not found, must specify index"));
```
wader, line 71: confusing with index? specify root id?

format/apple/bplist/bplist.md
dgmcdona: I like it, makes things more concise. I'll put that in

wader:
LGTM, tiny nitpick about error message. Let me know when if i should merge

btw have thought about renaming the current `from<format-name>` decode functions to `from_<format-name>` instead, think it makes it less cluttered. Currently each format has two functions, one named just the name that always returns a decode value and one with a `from`-prefix that throw error on decode error, not super happy by that design, a bit out of idea.

wader:
🎉

dgmcdona:
closes #502

wader:
Was thinking maybe add the `... (.. | .Bookmark? // empty) |= apple_bookmark` example to apple_bookmark doc?

PR #529, opened 2022-12-19
Add related file format projects to README

nichtich:


wader:
👍  I wonder if some formats in fq could have links to these projects? currently format refrences are kept in `<format-name>.md` ex https://github.com/wader/fq/blob/master/format/mp4/mp4.md#references

PR #612, opened 2023-03-09
readline: remove direct access to (*Instance).Config

slingamn:
I'm planning to remove `(*Instance).Config` as a public member in my fork [ergochat/readline](https://github.com/ergochat/readline), as part of a general project of shrinking the public API. This patch should be compatible with both chzyer/readline and the fork.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -35,6 +35,7 @@ func (a autoCompleterFn) Do(line []rune, pos int) (newLine [][]rune, length int)
 
 type stdOS struct {
 	rl            *readline.Instance
+	historyFile   string
 	closeChan     chan struct{}
 	interruptChan chan struct{}
 }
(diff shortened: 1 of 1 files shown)


wader:
Thanks! looks good, will merge once CI is happy

pkg/cli/cli.go
```diff
@@ -205,7 +207,7 @@ func (o *stdOS) Readline(opts interp.ReadlineOpts) (string, error) {
 
 func (o *stdOS) History() ([]string, error) {
 	// TODO: refactor history handling to use internal fs?
-	r, err := os.Open(o.rl.Config.HistoryFile)
+	r, err := os.Open(o.historyFile)
```
wader, line 210: Now seeing the TODO above i remember that it would be nice if readline had a history API that didn't use `os` directly. Would be more flexible and easier to test.

PR #646, opened 2023-04-23
readline: Switch to ergochat/readline

wader:


wader:
TODO
- bump config?
- undo in ergochat fork?
- expose term size/isterminal or just use same term module?

PR #650, opened 2023-04-30
bson: support all non-deprecated types and fix int/uint bugs

matthewdale:
* Add support for reading `javascript` elements.
* Add support for reading `decimal128` elements (only printing as binary for now, with a TODO to parse the decimal128 value).
* Add support for reading `minkey` and `maxkey` elements
* Fix BSON document length: read int32 instead of uint32.
* Fix binary element size: read int32 instead of uint32.
* Fix datetime: read int64 instead of int32.
* Update BSON test file to include at least one field for every supported type.


wader:
👍  Nice, will review later this evening

format/bson/bson.go
```diff
@@ -79,28 +86,38 @@ func decodeBSONDocument(d *decode.D) {
...
+						d.FieldUTF8NullFixedLen("value", int(length))
 					case elementTypeInt32:
 						d.FieldS32("value")
 					case elementTypeTimestamp:
 						d.FieldU64("value")
 					case elementTypeInt64:
 						d.FieldS64("value")
+					case elementTypeDecimal128:
+						// TODO: Parse the IEEE 754 decimal128 value.
+						d.FieldRawLen("value", 128)
```
wader, line 116: Would it make sense to decode but store it as float64 or just confusing? one could add `big.Rat` support to the decode API but not that useful as gojq dont' support it. Something to think about.

format/bson/testdata/test.fqtest
```diff
@@ -1,79 +1,175 @@
-# https://github.com/dwight/bsontools
-# echo '{array: [1,2,3], object: {key: "value"}, number: 123, string: "abc", true: true, false: false, null: null}' | fromjson test.bson
+# Run the code at https://go.dev/play/p/8YuESN04ACr and copy the base64 output. Then run:
+#
+#   echo "<base64 output>" | base64 -D > test.bson
+#
```
wader, line 4: I think it makes sense to add the generate program to a file under `format/bson/testdata`. Some reason it has to go thru base64 btw?

wader:
Looks good 👍  just some minor things about docs and tests

format/bson/testdata/test.fqtest
wader: Also thanks for good tests

format/bson/bson.go
matthewdale: I'd like to add support for decoding decimal128 values so the logical value can be displayed. I'm not sure yet how to integrate it into the `decode` API or the rest of `fq`, tho. Converting it to a float64 would be simple, but could be confusing because it might lose precision during conversion. I'll investigate further and see if I can come up with a good solution.

matthewdale: To clarify, I'm not currently planning to add decimal128 decoding support to this PR. Do you think it's OK to add decimal128 support in a subsequent PR?

format/bson/testdata/test.fqtest
wader: Aha makes sense

format/bson/bson.go
wader: No worries, and sorry that what i meant also, leave it for future improvement outside this PR. I can help out with adding `big.Rat` support if needed. One hacky way if you want to just show the value for now could be to assign it as a text description, hmm not great either :) but lets merge this without more feature and iterate on it if needed.

wader (approved):

wader:
Will merge once CI is green. Thanks a lot!

wader:
Fails on testing `fq -h bson`, `help(bson)`/`help("bson")` output:
```
--- FAIL: TestFormats (0.40s)
    --- FAIL: TestFormats/bson/testdata/help_bson.fqtest (0.03s)
        fqtest.go:15: 
            --- expected
            +++ actual
            @@ -8,6 +8,10 @@
               $ fq -d bson . file
               # Decode value as bson
               ... | bson
            +
            +Limitations:
            +
            +- The decimal1[28](https://github.com/wader/fq/actions/runs/4848437587/jobs/8641291596?pr=650#step:5:29) type is not supported for decoding, will just be treated as binary
             
             Convert represented value to JSON
             =================================
FAIL
```

Sorry I forgot in the review that `bson.md` should use markdown 3-level sections so change `Limitations:` to `### Limitations`.
After that run `WRITE_ACTUAL=1 go test -run TestFormats/bson ./format` and will rewrite the test output.

To be clear the markdown in `bson.md` is used both for `formats.md` and in the builtin help system where it's rendered a bit differently to be CLI friendly. Have some vague plan of using them to generate man pages etc also.

BTW if you want you can also add authors section, something like:
```
### Authors
- Mattias Wadman mattias.wadman@gmail.com, original author
- Matt Dale <contact info if you like>, additional types and bug fixes 
```

matthewdale:
@wader Thanks for the feedback and recommendations! I think I've made all the changes and fixes you suggested.

wader:
@matthewdale 🥳 Thanks and let me know if you have any question about the other improvements we talka about, happy to help out!

PR #669, opened 2023-05-13
repl: A bit nicer array completion

wader:


wader:
Or maybe should make it even more first hmm, only .[] if needed
```
            -frames[]
            +frames.[]
```

PR #671, opened 2023-05-15
bitcoin: fix witness item structs

Rogach:
This commit fixes witness item decoding such that they are represented as structs with `item_size` and `item` fields, instead of a flat list of raw bytes.


wader:
Hi thanks! do you have a small test case file you could add? it's usually just add the file to `bitcoin/testdata/<file>` add a `bitcoin/testdata/<file>.fqtest` with `$ fq -d <decoder if needed> dv <file>` and then run `WRITE_ACTUAL=1 go test -v -run TestFormats/bitcoin ./format/` to update expected output from actual.

Rogach:
Added a test.

wader:
Great, thanks! will merge once CI is green. There seems to be some hopefully temporary issue:
```
...
Err:1 http://azure.archive.ubuntu.com/ubuntu jammy/universe amd64 tcl-expect amd64 5.45.4-2build1
  503  Service Unavailable [IP: 40.81.13.82 80]
...
```

PR #673, opened 2023-05-15
interp: Add hex bits format

wader:
```
$ cat random.bin.gz | fq -rV -o bits_format=hex .uncompressed
f6f2074cf77d449d
```

Also made unknown bit formats an error.


wader:
@Rogach added hex bits format, see above for alternative way to output hex raw values, `-V` is for implicit `tovalue` to get JSON representation (or more correctly jq representation).

Feels like all the binary stuff needs a revisit and cleanup, bits_format should maybe be `binary_format` or `raw_format`? hmm

PR #707, opened 2023-06-20
sgpd box entries parsing

ksa-real:
Previously entries were parsed as raw bytes, now KIDs/IVs are explicitly parsed.


format/mp4/boxes.go
```diff
@@ -1367,27 +1367,48 @@ func decodeBox(ctx *decodeContext, d *decode.D, typ string) {
...
-			for i := uint64(0); i < entryCount; i++ {
-				entryLen := defaultLength
-				if version == 1 {
-					if defaultLength == 0 {
-						entryLen = d.FieldU32("description_length")
-					} else if entryLen == 0 {
-						d.Fatalf("sgpd groups entry len <= 0 version 1")
-					}
+
+		d.FieldStructNArray("entries", groupingType+" entry", int64(entryCount), func(d *decode.D) {
```
wader, line 1379: Most other boxes uses just entries/entry which i think is fine, don't think we need to repeat the type name. btw convention for the names are snake_case.

format/mp4/boxes.go
```diff
@@ -1367,27 +1367,48 @@ func decodeBox(ctx *decodeContext, d *decode.D, typ string) {
...
+					d.Fatalf("sgpd groups entry len == 0")
 				}
+			} else if entryLen == 0 {
+				// TODO: this is likely a mistake: here version is != 1, so defaultLength is default (i.e. 0),
+				// TODO: so entryLen is also 0. So for version != 1 this fatal should always throw
+				d.Fatalf("sgpd groups entry len == 0")
+			}
+			// CENC Sample Encryption Info Entries
+			switch groupingType {
+			case "seig":
```
wader, line 1394: Have any small example file with this? would be great with a test if possible

wader:
Thanks! hade some minor comments

format/mp4/boxes.go
ksa-real: The issue was that there are two different box types depending on `groupingType`. The intent was to indicate it somehow, but I'll just do the common way.

format/mp4/boxes.go
ksa-real: I noticed you were using shaka packager "dd987007-release". The official release is pretty outdated (2001). Were you able to build from source or there is a more recent release somewhere?

wader: Hmm 🤔 not sot sure how i ran it to produce those files, it could have been that i had "packager" as an shell alias for running shaka-packager via docker and that one was old. Seems like those tests were added in 2021 161dcaffd4e1d73fb4e25015a4385796d42e35ee

Anyways, it's probably a good idea that we regenerate with new version 👍

wader:
LGTM, thanks!

PR #709, opened 2023-06-20
luajit: add decoder

dlatchx:
a decoder for LuaJIT bytecode/precompiled files


format/luajit/luajit.go
```diff
@@ -0,0 +1,355 @@
...
+	BigEndian bool
+}
+
+func LuaJITDecodeHeader(di *DumpInfo, d *decode.D) {
+	magic := d.FieldStr("magic", 3, encoding.Nop)
+
+	ok := bytes.Equal([]byte(magic), []byte{0x1b, 0x4c, 0x4a})
+	if !ok {
+		d.Errorf("invalid magic number")
+	}
```
wader, line 67: Think you can replace this with `d.FieldStr("magic", 3, d.AssertStr(".."))` or `d.FieldRawLen("magic", 3*8, d.AssertBitBuf([]byte{...}))` which will do the compare and also the magic will be shown as "valid"

format/luajit/luajit.go
```diff
@@ -0,0 +1,355 @@
...
+	d.FieldU8("version")
+
+	var flags uint64
+	d.FieldStruct("flags", func(d *decode.D) {
+		flags = d.FieldULEB128("raw")
+
+		d.FieldValueBool("BE", flags&0x01 > 0)
+		d.FieldValueBool("STRIP", flags&0x02 > 0)
+		d.FieldValueBool("FFI", flags&0x04 > 0)
+		d.FieldValueBool("FR2", flags&0x08 > 0)
```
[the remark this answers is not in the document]
wader: Can also use `0b...` notation if you want to be explicit about bit positions

format/luajit/luajit.go
```diff
@@ -0,0 +1,355 @@
...
+	})
+
+	di.Strip = flags&0x2 > 0
+	di.BigEndian = flags&0x1 > 0
+
+	if !di.Strip {
+		namelen := d.FieldU8("namelen")
+		d.FieldStr("name", int(namelen), encoding.Nop)
```
wader, line 86: Is `encoding.Nop` important? can skip if it's ok to be decoded as UTF-8

format/luajit/opcodes.go
```diff
@@ -0,0 +1,175 @@
...
+func (op *BcDef) IsJump() bool {
+	return op.MC == BcMjump
+}
+
+var bcOpSyms = scalar.UintMapSymStr{}
+
+func init() {
+	for i := 0; i < len(opcodes); i++ {
+		bcOpSyms[uint64(i)] = opcodes[i].Name
+	}
```
wader, line 173: Wonder if you can make the `BcDef` implement `scalar.UintMapper`, i think the `formatEntries` type in `msgpack.go` does something similar. If so you can skip init and `bcOpSyms`

wader (commented):
Thanks! Did some quick review and will do some more later. I think we should be able to add some custom decode function to replace some FieldAnyFn and maybe `kgcType` could use some standard mapper?

Overall looks good i think and thanks for including tests!

BTW you can add a `luajit.md` with author add embedd it, see wasm.go/wasm.md for example. Then it will be part of the documentation and also available in the CLI help.

format/luajit/luajit.go
```diff
@@ -0,0 +1,358 @@
...
+)
+
+//go:embed luajit.md
+var LuaJITFS embed.FS
+
+func init() {
+	interp.RegisterFormat(
+		format.LuaJIT,
+		&decode.Format{
+			Description: "LuaJIT 2.0 bytecode dump",
```
wader, line 41: Possibly skip the "dump" part or are .luac files known as dump files?

dlatchx:
> I think we should be able to add some custom decode function to replace some FieldAnyFn
I added more decode function, but is still need FieldAnyFn, because LuaJIT's format makes a weird use of ULEB128 encoding, and I want to map the bits of the two 32bit-max ULEB128 to the resulting 64bit Field. (see LuaJITDecodeI64(), for example)
Maybe there's another way to do that ?

> ... and maybe `kgcType` could use some standard mapper?

The problem is that we need a default case: 0 to 4 are types with know size, but 5 and more indicates a string *and* its size:
5 is a 0-length string, 6 is a 1-byte string, ect.

I don't think we can use a UintMapSymStr here.

dlatchx:
> ... and maybe `kgcType` could use some standard mapper?

The problem is that we need a default case: 0 to 4 are types with know size, but 5 and more indicates a string *and* its size:
5 is a 0-length string, 6 is a 1-byte string, ect.

I don't think we can use a UintMapSymStr here.

wader:
> I added more decode function, but is still need FieldAnyFn, because LuaJIT's format makes a weird use of ULEB128 encoding, and I want to map the bits of the two 32bit-max ULEB128 to the resulting 64bit Field. (see LuaJITDecodeI64(), for example) Maybe there's another way to do that ?

I see, yeap that might be tricky.

>The problem is that we need a default case: 0 to 4 are types with know size, but 5 and more indicates a string and its size:
> 5 is a 0-length string, 6 is a 1-byte string, ect.
>
> I don't think we can use a UintMapSymStr here.

Aha that makes sense, i guess you could  use a range mapper but maybe not worth it.

I added some more comments but getting a bit late here so will continue tomorrow. Think we're close to something mergable. Thanks for the patience!

format/luajit/luajit.go
```diff
@@ -0,0 +1,381 @@
...
+var probeGroup decode.Group
+
+func init() {
+	interp.RegisterFormat(
+		format.LuaJIT,
+		&decode.Format{
+			Description: "LuaJIT 2.0 bytecode",
+			Groups:      []*decode.Group{format.Probe},
+			DecodeFn:    LuaJITDecode,
+			Dependencies: []decode.Dependency{
```
wader, line 46: `Dependencies` you can skip, it's if the format decoder itself want to decode something using the probe group

format/luajit/luajit.go
```diff
@@ -0,0 +1,381 @@
...
+			})
+
+			d.FieldArray("knum", func(d *decode.D) {
+				for i := uint64(0); i < numkn; i++ {
+					d.FieldAnyFn("knum", LuaJITDecodeKNum)
+				}
+			})
+
+			if !di.Strip {
+				d.FieldArray("debug", func(d *decode.D) {
```
wader, line 341: When i played around with the decoder i noticed that data in debug has some strings, is there any known structure to it or should be threaded as opaque blob? if blob should the array be replaced with `d.FieldRawLen("debug", debuglen*8)`?

format/luajit/luajit.go
```diff
@@ -0,0 +1,381 @@
...
+	lo := d.ULEB128()
+	if lo&1 == 0 {
+		return lo >> 1
+	} else {
+		hi := d.ULEB128()
+		return u64tof64((hi << 32) + (lo >> 1))
+	}
+}
+
+type kgcType struct{}
```
wader, line 176: One varint is to do something like:
```go
type fallbackUintMapSymStr struct {
	fallback string
	scalar.UintMapSymStr
}

func (m fallbackUintMapSymStr) MapUint(s scalar.Uint) (scalar.Uint, error) {
	s.Sym = m.fallback
	return m.UintMapSymStr.MapUint(s)
}
```
and then
```go
	kgctype := d.FieldULEB128("type", fallbackUintMapSymStr{
		fallback: "str",
		UintMapSymStr: scalar.UintMapSymStr{
			0: "child",
			1: "tab",
			2: "i64",
			3: "u64",
			4: "complex",
		},
	})
```
and use it for `ktabType` also

format/luajit/luajit.go
```diff
@@ -0,0 +1,381 @@
...
+			d.FieldStruct("khash", func(d *decode.D) {
+				d.FieldStruct("k", LuaJITDecodeKTabK)
+				d.FieldStruct("v", LuaJITDecodeKTabK)
+			})
+		}
+	})
+}
+
+func LuaJITDecodeI64(d *decode.D) {
+	d.FieldAnyFn("i64", func(d *decode.D) any {
```
wader, line 217: For some other format decoders that do something similar i've decode the value as a "value" field, see msgpack or bson etc. Maybe that would make sense here also? for complex maybe the value field could an extra `d.FieldValueAny("value", ...)`?

format/luajit/luajit.go
```diff
@@ -0,0 +1,381 @@
...
+			Groups:      []*decode.Group{format.Probe},
+			DecodeFn:    LuaJITDecode,
+			Dependencies: []decode.Dependency{
+				{Groups: []*decode.Group{format.Probe}, Out: &probeGroup},
+			},
+		})
+	interp.RegisterFS(LuaJITFS)
+}
+
+// reinterpret an int as a float
```
wader, line 53: Trying to understand how this works. So all numbers are floats but encoded i two parts where the hi part will be zero for < 32 bit integers? and hi might be > 2^32 so hi << 32 + lo will be >  2^64 and needs to be float? but looks like we only care about 8 bytes when reinterpreting? hope that made sense :)

format/luajit/luajit.go
```diff
@@ -0,0 +1,409 @@
...
+		// drop the LSB
+		lo_data := lo >> 1
+
+		// downcast to 32bits. should not overflow
+		lo_data_u32 := uint32(lo_data)
+
+		// make it a signed integer
+		lo_data_i32 := int32(lo_data_u32)
+
+		// return a larger type to make fq happy
```
wader, line 285: 🕺💃  maybe there should be smaller number types hmm, should probably do another round of profiling memory soon

format/luajit/luajit.go
```diff
@@ -0,0 +1,409 @@
...
+func LuaJITDecodeDebug(d *decode.D, debuglen uint64, numbc uint64) {
+	d.FieldStruct("debug", func(d *decode.D) {
+		d.FieldArray("lines", func(d *decode.D) {
+			for i := uint64(0); i < numbc; i++ {
+				d.FieldU8("value")
+			}
+		})
+
+		// TODO: find out more about how to decode these strings
+		d.FieldArray("anotations", func(d *decode.D) {
```
wader, line 304: typo: annotations

👍 with more debug structure

wader:
Lint not happy about unused `probeGroup`

Think i'm happy to merge if your happy too?

dlatchx:
it's good for me

wader:
Thanks!

PR #747, opened 2023-08-18
moc3 and caff: add decoder

Ronsor:
This PR adds support for the proprietary MOC3 and CAFF formats used by the Live2D Cubism software (https://live2d.com/en/).

Notes:
* MOC3 is a format for 2D rigged puppets, somewhat like Flash.
* CAFF is an archive format usually found with the extensions `.cmo3` and `.can3`.
* No test data is included since there isn't any that I'm allowed to redistribute but can be found at https://www.live2d.com/en/download/sample-data/.
* The MOC3 decoder is based on the specification I wrote: https://github.com/OpenL2D/moc3ingbird/blob/master/src/moc3.hexpat


format/caff/caff.go
```diff
@@ -0,0 +1,205 @@
...
+	d.ArgAs(&ci)
+
+	var obfsKey uint64
+
+	obfsU8 := func(d *decode.D) uint64 { return d.U8() ^ (obfsKey & 0xFF) }
+	obfsU32 := func(d *decode.D) uint64 { return d.U32() ^ (obfsKey & 0xFFFFFFFF) }
+	obfsU64 := func(d *decode.D) uint64 { return d.U64() ^ (obfsKey<<32 | obfsKey) }
+	obfsBool := func(d *decode.D) bool { return obfsU8(d) != 0 }
+
+	// "Big Endian Base 128" - LEB128's strange sibling
```
wader, line 96: 😄 i guess strange enough to not include in `*decode.D`?

format/moc3/moc3.go
```diff
@@ -0,0 +1,1646 @@
...
+
+func decodeMOC3(d *decode.D) any {
+	d.FieldUTF8("magic", 4, d.StrAssert("MOC3"))
+	version := d.FieldU8("version", moc3VersionNames)
+	isBigEndian := d.FieldBoolFn("is_big_endian", func(d *decode.D) bool { return d.U8() != 0 })
+
+	if !isBigEndian {
+		d.Endian = decode.LittleEndian
+	}
+	d.SeekRel(58 * 8)
```
wader, line 354: Does this cause "unknown" fields? one way i've done is to explicitly add something like "unused" fields for bit ranges that i known to be unused. "unknown" fq adds automatically for ranges no field as "claimed"

format/caff/caff.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func decodeCAFF(d *decode.D) any {
+	var ci format.CAFF_In
+	d.ArgAs(&ci)
+
+	var obfsKey uint64
+
+	obfsU8 := func(d *decode.D) uint64 { return d.U8() ^ (obfsKey & 0xFF) }
+	obfsU32 := func(d *decode.D) uint64 { return d.U32() ^ (obfsKey & 0xFFFFFFFF) }
```
wader, line 92: No a big thing but rest of the code uses lowercase and grouping for hex, binary literals

wader (commented):
Had a quick look and think it looks great overall 👍 had some minor comments. I will take a deeper look and try the decoder myself later today.

BTW do you know if there are MIT licenses etc model files somewhere we can use for tests? or could we alternatively create some simple models in live2d and save and use those? the reason i really want test files is to be able to refactors and do changes with some kind of insurance to not break or change things in a bad way.

wader:
The i386 failure is just about test for probe order and help text is failing. Should be fixable with `WRITE_ACTUAL=1 go test ./format` and commit the changes to those tests

format/caff/caff.go
Ronsor: It could be, but dealing with the obfuscation makes it awkward and I don't know of any other format that uses it.

Richardn2002:
https://github.com/RavioliMavioli/archlive2d
This is a repo containing one moc3 file under the CC0 1.0 license.

https://github.com/shuhaiwen/live2dw
This is a repo containing tons of live2d models but:

0. The author does not specify the sources clearly, nor specify a license.
1. They are moc instead of moc3 files.

And I have to admit that I am not involved in the development of this PR, nor have experiences in the moc3 format, which means I do not know if these models are enough for testing.

wader:
@Richardn2002 Hey! thanks that is good pointers. Ideally for tests some kind of simplest possible files that still use lots of features is best.

Maybe easiest is to download live2d, run it and save some simple files?

Richardn2002:
Yeah, would be nice to use Live2D with all functionalities possible to create a test model. I sadly have not purchased Live2D Cubism. I hope @Ronsor has one copy.

wader:
Yeap would be great, but i see there is a free for 42 days version 🤔

format/caff/caff.go
Ronsor: Updated for consistency 👍

format/moc3/moc3.go
```diff
@@ -0,0 +1,1646 @@
...
+	moc3Version3_03_00 = 2
+	moc3Version4_00_00 = 3
+	moc3Version4_02_00 = 4
+)
+
+var moc3VersionNames = scalar.UintMap{
+	moc3Version3_00_00: {Sym: "3_00_00", Description: "3.0.00 - 3.2.07"},
+	moc3Version3_03_00: {Sym: "3_03_00", Description: "3.3.00 - 3.3.03"},
+	moc3Version4_00_00: {Sym: "4_00_00", Description: "4.0.00 - 4.1.05"},
+	moc3Version4_02_00: {Sym: "4_02_00", Description: "4.2.00 - 4.2.02"},
```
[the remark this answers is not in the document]
Ronsor: I mainly chose the `V3_00_00`, etc. strings to match the enum from the spec.

Ronsor:
> https://github.com/shuhaiwen/live2dw This is a repo containing tons of live2d models but:
> 
> 0. The author does not specify the sources clearly, nor specify a license.
> 1. They are moc instead of moc3 files.

Besides the old format, they seem to be taken from Live2D's previous sample downloads page.

format/caff/caff.go
wader: agree, lets keep it here

Ronsor:
I made some improvements:

* Included CAFF test data (test.cmo3)
* Fixed bug with CAFF deobfuscation
* Updated the MOC3 specification to add the new fields for version 5 (released a couple days ago).
* Updated the MOC3 decoder to reflect this.
* Included MOC3 test data (based on the Arch Chan MOC3, converted to version 5)
* Improved generated tree structure for MOC3 with better names for array elements and grouping section fields together in spite of version

format/caff/caff.go
```diff
@@ -0,0 +1,210 @@
...
+	d.FieldStruct("archive_version", decodeVersion)
+	d.FieldUTF8("format_id", 4)
+	d.FieldStruct("format_version", decodeVersion)
+	obfsKey = d.FieldS32("obfuscate_key")
+	d.SeekRel(8 * 8)
+
+	d.FieldStruct("preview_image", func(d *decode.D) {
+		d.FieldU8("image_format", imageFormatNames)
+		d.FieldU8("color_type", colorTypeNames)
+		d.SeekRel(2 * 8)
```
wader, line 131: Could this and some of the other `d.SeekRel` be a unused/alignment/padding etc field? i usually try to have no or as few gaps as possibly but i understand it might hard when reversing a format.

BTW when i was talking about "unknown" fields earlier i was confused, they called "gap" fields now, was renamed some time ago so that decoders can use "unknown".

format/caff/caff.go
```diff
@@ -0,0 +1,210 @@
...
+				d.FieldValueUint("compress_option", uint64(entry.compressOption), compressOptionNames)
+
+				rawBytes := d.BytesLen(entry.fileSize)
+				if entry.isObfuscated {
+					for i, v := range rawBytes {
+						rawBytes[i] = v ^ uint8(obfsKey)
+					}
+				}
+
+				br := bitio.NewBitReader(rawBytes, -1)
```
wader, line 176: Sorry for confusing APIs for doing this atm. Hopefully in the future there could be some helpers for dealing with compressed/massaged data

format/moc3/moc3.go
```diff
@@ -0,0 +1,1766 @@
...
+		targetIndices                               int64
+		blendShapeKeyformBindingSourcesBeginIndices int64
+		blendShapeKeyformBindingSourcesCounts       int64
+	}
+}
+
+func decodeMOC3(d *decode.D) any {
+	d.FieldUTF8("magic", 4, d.StrAssert("MOC3"))
+	version := d.FieldU8("version", moc3VersionNames)
+	isBigEndian := d.FieldBoolFn("is_big_endian", func(d *decode.D) bool { return d.U8() != 0 })
```
wader, line 372: Maybe want to use sym map to a bool? i usually try to make actual the decoded value and then sym the interpretation of the value. That way a user has access to both (`toactual`/`tosym`, default is sym valie if set, otherwise actual)

format/moc3/moc3.go
```diff
@@ -0,0 +1,1766 @@
+package moc3
+
+// https://github.com/OpenL2D/moc3ingbird/blob/master/src/moc3.hexpat
```
wader, line 3: 👍 interesting to see two implementations, are the comparable completnesswise? and as someone who has worked with both, how do they compare? the go code feels a bit more verbose?

format/moc3/moc3.go
```diff
@@ -0,0 +1,1766 @@
...
+	}
+	d.FieldRawLen("unused", 58*8)
+
+	var sectionOffsets sectionOffsetTable
+	d.FramedFn(0x280*8, func(d *decode.D) {
+		d.FieldStruct("section_offsets", func(d *decode.D) {
+			sectionOffsets.countInfo = int64(d.FieldU32("count_info"))
+			sectionOffsets.canvasInfo = int64(d.FieldU32("canvas_info"))
+
+			d.FieldStruct("parts", func(d *decode.D) {
```
wader, line 385: This code makes me think that maybe the decode API should have something similar to `Read` in `encoding/binary` hmm

wader (commented):
Nice work on this, already looked good and gets better, have a look at my comments.

Would be great to cut down a bit on gap fields, make into explicit unknown/reserved/align/padding and whatnot is possible.

And great with test files, will make future maintenance much easier. Are they ok licensewise? test files are not shipped in the binary but i guess its good to keep the source distribution ok licensewise also.

format/caff/caff.go
```diff
@@ -0,0 +1,210 @@
...
+			raw[i] ^= byte(obfsKey)
+		}
+		return string(raw)
+	}
+
+	d.FieldUTF8("archive_id", 4, d.StrAssert("CAFF"))
+	d.FieldStruct("archive_version", decodeVersion)
+	d.FieldUTF8("format_id", 4)
+	d.FieldStruct("format_version", decodeVersion)
+	obfsKey = d.FieldS32("obfuscate_key")
```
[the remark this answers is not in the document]
Ronsor: I found out the hard way that it needs to be signed otherwise `obfsU64()` can return bad data.

format/caff/caff.go
Ronsor: I'll change those to `unusedN` fields since I know they're unused here.

format/moc3/moc3.go
Ronsor: The Go code is largely autogenerated from the original implementation and then edited. They're at feature parity.

Some of the verbosity of the go decoder is probably due to some code duplication I can fix.

Ronsor:
> Are they ok licensewise?

The MOC3 test is CC0 data, the CAFF test is my own work. Licensing should be no problem.

format/moc3/moc3.go
Ronsor: `FieldUnmarshalStruct(field string, out any)`?

format/caff/caff.go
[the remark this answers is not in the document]
wader: Aha i see 👍

format/moc3/moc3.go
wader: Yeap something like that. Auto camel_case names and maybe use struct tags to support some casting. Something for the future, no need for this PR.

wader:
> > Are they ok licensewise?
> 
> The MOC3 test is CC0 data, the CAFF test is my own work. Licensing should be no problem.

Great!

Ronsor:
@wader What do you think is the best way to add alignment for arrays (at the end, not per field)?

format/moc3/moc3.go
```diff
@@ -0,0 +1,1766 @@
...
+			d.FieldU8("canvas_flags")
+		})
+
+		d.FieldStruct("parts", func(d *decode.D) {
+			d.SeekAbs(sectionOffsets.parts.ids * 8)
+			d.FieldArray("ids", func(d *decode.D) {
+				for i := int64(0); i < countInfo.parts; i++ {
+					d.FieldUTF8NullFixedLen("id", 64)
+				}
+			})
```
wader, line 732: @Ronsor you mean alignment here for example? easiest is probably just an extra "ids_alignment" etc field after, other maybe more correct but less user friendly is to make "ids" into struct with "entries" and "alignment" i think

Ronsor: Yes, there. I'll try both ways to see which is more usable.

Ronsor:
I settled on making each section a struct with both `array` and `padding` fields, and I refactored the code to be a lot less repetitive. Also fixed an issue with decoding version 5 files (the correction is reflected in the specification).

wader:
LGTM! Think i'm ready to merge this once CI is green. Do you have something more you want to change or add?

The CI failure seems to be just some field that has been renamed

Ronsor:
Whoops, forgot to update the test. I think it's ready to be merged now. 🥳

wader:
Thanks! Great work and hope you find fq usefull! any plans for future contributions?

Ronsor:
Right now there's about three other proprietary formats I plan to add decoders for in the near future.

wader:
> Right now there's about three other proprietary formats I plan to add decoders for in the near future.

Interesting! looking forward

BTW forgot to say that you can add documentation and authorship to a decoder by doing something like https://github.com/wader/fq/blob/master/format/apple/macho/macho.md and https://github.com/wader/fq/blob/master/format/apple/macho/macho.go#L17-L28 and it will show up in cli help and `make doc` will generate documentation.

Issue #697, opened 2023-06-11
[Documentation] Any interest in creating a man page?


zarakshR:
Would there be any interest in writing a man page for fq? I would love to help as far as possible.

wader:
Hey! yes that would be nice and i've done some experiments how to do it but nothing complete or decided yet.

I just pushed https://github.com/wader/fq/pull/698 with one way of doing it. I guess it depends a bit what we want to have in the man page, list for formats? my idea with https://github.com/wader/fq/pull/698 is to try share most of the content between cli, documentation and man page so it's less work to maintain.

What do you think? any help is appreciated! i'm new to how to write good and compatible (linux and BSD) man pages, so that would be very helpful.

zarakshR:
Hi! Sorry for the late reply, was caught up in school work.

The `man(7)` man-page lays out how to create man page for linux (I cannot run my FreeBSD system at the moment so not sure about how man works over there) The man command basically runs `groff` in the background with a special macro template called `an` and displays `groff`'s output to the screen.

`man(7)` also lays out the basic structure required to define a man documentation page.

[Here](https://pasteboard.co/y58TWk0ogKth.png) is a basic example of how it works. The window on the left is the source file which `groff` (called by `man`) processes into the output on the left.

I think I could write up a man source file pretty quickly, but where should I get the info from? Do I just get it from the README or do you have any other documentation source somehwere?

wader:
Hey, no worries. As i noted in the comment above it would be nice to be able to share/generate part of the man page. Ideally i think i would like to generate the man page based on some sections from README.md and also generate support formats, options, env variables etc. In https://github.com/wader/fq/pull/698 i'm did a stab at having a jq script generate the man page, fq also has support for markdown so it should be possible to parser README.md and pick things out.

But maybe a first step could be to just have something static and start with that? maybe just copy text from README.md? that would help to have as base for for a script the generate.

zarakshR:
OK. I am starting work on writing a static version of the man page. I'll use the README.md and the usage.md as a starting point, is that OK?

P.S. Maybe look into using `m4`, the macro processor, it would probably simplify dynamically generating different kinds of documentations/accreditation/upstream contact information/maintainers list/etc

wader:
Sounds great 👍 can also sneak peak at the diff for https://github.com/wader/fq/pull/698

PR #764, opened 2023-09-08
doc: Add man page

zarakshR:
This implements a man page as in #697

The man page file is `doc/fq.1` formatted as a `troff` file using the man-pages project default man page template.

You can get a look at how the final man page will run by using `man -l doc/fq.1`

### The structure of the man page is as follows -

NAME

SYNOPSIS

DESCRIPTION

INVOKING FQ
    Command Line Arguments
        * Decoding/Encoding Options (-o options)
    Supported Formats

CONFIGURATION

THE FQ LANGUAGE
    TYPES SPECIFIC TO FQ
        Decoded Values
        Binary
        Binary Array
    FUNCTIONS
        Functions Added in fq
        Naming Inconsistencies
    ENCODING, SERIALIZATIONS, HASHES
    USE AS SCRIPT INTERPRETER

BUGS (+ KNOWN ISSUES)

AUTHOR

SEE ALSO

### Some notes - 

"Use representation of a format" not added to examples -- not sure how I could fit it in

Have not included the details of each format (`formats.md`)

The section "Display Output" has been merged into "Functions"

The examples for each of the types in "types specific to fq" were not added -- not sure how I should format them

Only the first paragraph of "Encodings, Serializations, Hashes" was added; I couldn't figure out how to structure them

"Own decoders and use as library" section is TODO in usage.md and was not included

### DIFF
--- a/doc/fq.1
+++ b/doc/fq.1
@@ -6,6 +6,24 @@ fq \- jq for binary formats
 .SH SYNOPSIS
 \fBfq\fR [\fI\,OPTION\/\fR]... [--] \fI\,EXPR [\fI\,FILE\/\fR]...
 
+\fBfq\fR . <FILE>
+
+\fBfq\fR d <FILE>
+
+\fBfq\fR -V '.path[1].value' <FILE>
+
+\fBfq\fR tovalue <FILE>
+
+\fBfq\fR -r to_toml file.yml
+
+\fBfq\fR -s -d html 'map(.html.head.title?)' *.html
+
+\fBfq\fR file.cbor | fq -d cbor torepr
+
+\fBfq\fR 'grep("^main$") | parent' /bin/ls
+
+\fBfq\fR -i
+
 .SH DESCRIPTION
 \fBfq\fR is inspired by the well known jq tool and language that allows you to work with binary formats the same way you would using jq. In addition it can present data like a hex viewer, transform, slice and concatenate binary data. It also supports nested formats and has an interactive REPL with auto-completion.
 .P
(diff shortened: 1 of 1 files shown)


doc/fq.1
```diff
@@ -0,0 +1,392 @@
...
+Set a decoding/encoding option. A list of supported options are -
+.RS
+.P
+\fBbits_format\fR - string, md5, hex, base64, truncate, snippet
+.RS
+Controls the representation of raw binary in the output JSON
+.RE
+.RE
+.SS Supported Formats
+Use \fBfq --help formats\fR to get a list of supported formats
```
wader, line 53: If we're nog going to auto generate this man page maybe this could instead give example how to list supported formwats?

doc/fq.1
```diff
@@ -0,0 +1,392 @@
+.TH FQ "1" "July 2023"
+
+.SH NAME
+fq \- jq for binary formats
+
+.SH SYNOPSIS
+\fBfq\fR [\fI\,OPTION\/\fR]... [--] \fI\,EXPR [\fI\,FILE\/\fR]...
```
wader, line 7: It would be nice to have some usage examples here i think. maybe same as shown for fq -h?

doc/fq.1
zarakshR: I'm not sure I understand this question, isn't `fq --help formats` the way to list supported formats?

wader: Ah sorry i totally missed that, only saw the list of formats :) and what i was wondering was if we should include the list in the man page or not as it will have to be manually kept up to date if we don't update/generate it with `make doc` etc.

PR #766, opened 2023-09-19
moc3: Fix minor bugs

Ronsor:
This PR simply fixes version detection for the `count_info` structure (`glue_keyforms` was previously incorrectly restricted to format version 4.2.00+) and the order of fields in `blend_shape_keyform_bindings` (`keyform_sources_*` fields should be before `blend_shape_constraint_*` fields). The tests have been updated to reflect this also.


wader:
Thanks!

PR #769, opened 2023-09-26
add OpenTimestamps file parser

fiatjaf:
https://opentimestamps.org/ (there is no spec or docs about the file format in this website, only libraries).

This is my first time contributing and I don't know if I did this right. I did read the developer guide.

One thing I wanted to know is how to make `.ots` files be detected automatically. They have this clear extension in their filenames and also an obvious sequence "magic bytes" at their beginning, so it should be pretty easy to detect without having to parse them entirely.


format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	return value
+}
+
+var (
+	headerMagic  = []byte{0x00, 0x4f, 0x70, 0x65, 0x6e, 0x54, 0x69, 0x6d, 0x65, 0x73, 0x74, 0x61, 0x6d, 0x70, 0x73, 0x00, 0x00, 0x50, 0x72, 0x6f, 0x6f, 0x66, 0x00, 0xbf, 0x89, 0xe2, 0xe8, 0x84, 0xe8, 0x92, 0x94}
+	pendingMagic = binary.BigEndian.Uint64([]byte{0x83, 0xdf, 0xe3, 0x0d, 0x2e, 0xf9, 0x0c, 0x8e})
+	bitcoinMagic = binary.BigEndian.Uint64([]byte{0x05, 0x88, 0x96, 0x0d, 0x73, 0xd7, 0x19, 0x01})
```
wader, line 43: Could be written as `bitcoinMagic uint64 = 0x05_0x88_0x96_0x0d_0x73_0xd7_0x19_0x01` i think?

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/wader/fq/pkg/interp"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+func init() {
+	interp.RegisterFormat(
+		format.OpenTimestamps,
+		&decode.Format{
+			Description:  "OpenTimestamps file",
+			Dependencies: nil,
```
wader, line 19: To make the format probeable add it to the probe group `Groups: []*decode.Group{format.Probe},`. fq by design currently do not have any special per format code for probing it's up to a format decoder to fail early which seems to be the case here

format/format.go
```diff
@@ -83,6 +83,7 @@ var (
 	Bitcoin_Block       = &decode.Group{Name: "bitcoin_block"}
 	Bitcoin_Script      = &decode.Group{Name: "bitcoin_script"}
 	Bitcoin_Transaction = &decode.Group{Name: "bitcoin_transaction"}
+	OpenTimestamps      = &decode.Group{Name: "ots"}
```
wader, line 86: I prefer to keep the names in sync so could the name `open_timestamps`?

format/opentimestamps/ots.go
wader, line 19: Can remove the `Dependencies: nil,` line

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+var (
+	headerMagic  = []byte{0x00, 0x4f, 0x70, 0x65, 0x6e, 0x54, 0x69, 0x6d, 0x65, 0x73, 0x74, 0x61, 0x6d, 0x70, 0x73, 0x00, 0x00, 0x50, 0x72, 0x6f, 0x6f, 0x66, 0x00, 0xbf, 0x89, 0xe2, 0xe8, 0x84, 0xe8, 0x92, 0x94}
+	pendingMagic = binary.BigEndian.Uint64([]byte{0x83, 0xdf, 0xe3, 0x0d, 0x2e, 0xf9, 0x0c, 0x8e})
+	bitcoinMagic = binary.BigEndian.Uint64([]byte{0x05, 0x88, 0x96, 0x0d, 0x73, 0xd7, 0x19, 0x01})
+)
+
+var attestationMapper = scalar.UintMapSymStr{
+	pendingMagic: "calendar",
```
wader, line 47: Should this be `pendingMagic: "pending"`? but i know nothing about ots :)

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+func decodeOTSFile(d *decode.D) any {
+	d.Endian = decode.BigEndian
+
+	d.FieldRawLen("magic_bytes", int64(8*len(headerMagic)), d.AssertBitBuf(headerMagic))
+	d.FieldUintFn("version", decodeVarInt)
+	tag := d.FieldU8("digest_hash_algorithm", opMapper,
+		scalar.UintDescription("algorithm used to hash the source file"))
+	if tag != 8 {
+		name := opMapper[tag]
+		d.Errorf("only sha256 supported, got %x: %s", tag, name)
```
wader, line 73: If there no point of continue decode you can use `d.Fatalf`. `d.Errorf` is possible to ignore with `-o force=true`

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil
+	}
+	d.FieldRawLen("digest", 8*32, scalar.RawHex,
+		scalar.BitBufDescription("hash of the source file"))
+
+	d.FieldArray("instructions", func(d *decode.D) {
+		for {
+			if b, err := d.TryPeekBytes(1); errors.Is(err, io.EOF) {
```
wader, line 81: Could this be replace with:
```go
for !d.End() {
        b := d.PeekUintBits(8)
        // ...
}   
```

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+				break
+			} else if b[0] == 0x00 {
+				d.FieldStruct("attestation", func(d *decode.D) {
+					d.FieldU8("attestation_tag", scalar.UintMapSymBool{0x00: true})
```
wader, line 85: Not sure i follow what the uint to bool mapping i meant to do here and below?

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+					case pendingMagic:
+						nurl := d.FieldUintFn("url_size", decodeVarInt)
+						d.FieldUTF8("url", int(nurl),
+							scalar.StrDescription("url of the calendar server to get the final proof"))
+					default:
+						d.Errorf("unknown attestation tag %x", val)
+					}
+				})
+			} else if b[0] == 0xff {
+				d.FieldStruct("continuation_byte", func(d *decode.D) {
```
wader, line 101: Is the struct needed here? could be just a `d.Field8`?

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+					d.FieldU8("continuation_byte", scalar.UintMapSymBool{0xff: true},
+						scalar.UintDescription("tells we should continue reading after the next attestation block"))
+				})
+			} else {
+				d.FieldStruct("instruction", func(d *decode.D) {
+					tag := d.FieldU8("op", opMapper)
+					if name, ok := opMapper[tag]; ok {
+						// read var bytes if argument
+						if name == "append" || name == "prepend" {
```
wader, line 110: In other decode i've usually preferred to use constant, something like this https://github.com/wader/fq/blob/master/format/bson/bson.go#L29-L69 but i guess you would still have to lookup the tag in opMapper unless you do some kind of switch with tags that has no args, has args and a default case with error

format/opentimestamps/testdata/hello-world.fqtest
```diff
@@ -0,0 +1,1177 @@
+$ fq -d ots dv hello-world.txt.ots
```
wader, line 1: Once in the probe group i think you can skip `-d` so that the probing is tested also

wader:
> https://opentimestamps.org/ (there is no spec or docs about the file format in this website, only libraries).

Your welcome to add URLs to code etc you used as reference to the top of ots.go or to a ots.md file, also you might want a ots.md file in any case, example file https://github.com/wader/fq/blob/master/format/bson/bson.md and see bson.go about embedding it to have cli help also.

And you can use `make doc` to regenerate part of the documentation to include your format.

> This is my first time contributing and I don't know if I did this right. I did read the developer guide.

🥳 hope it gave some good tips at least

> One thing I wanted to know is how to make `.ots` files be detected automatically. They have this clear extension in their filenames and also an obvious sequence "magic bytes" at their beginning, so it should be pretty easy to detect without having to parse them entirely.

I add a code comment about how automatic detection works. It's is currently solely done on content, maybe in the future a decoder will get access to the filename somehow

BTW the CI error seems to be that a common format test i failing, can updated using `WRITE_ACTUAL=1 go test -run TestFormats/all/all.fqtest ./format`

format/format.go
fiatjaf: Can it be `opentimestamps`?

format/opentimestamps/ots.go
fiatjaf: Yes, I don't know either. The thing is that we're matching on the `0x00` already to determine this is an attestation block -- then we must turn that same byte (that we already know because we peeked into it) into a named field so `fq` doesn't think it's a gap, right? So I thought having a `"attestation_tag": true` would be the least bad way of doing this, but I don't know. What is the best approach?

format/opentimestamps/ots.go
fiatjaf: I initially wrote it with a `FieldU8`, but then it doesn't show a name in the output, it just shows `255` in the list.

Should I perhaps translate that into a string?

format/opentimestamps/ots.go
wader: I see, maybe can be modelled like this?
```go
d.FieldArray("instructions", func(d *decode.D) {
        for !d.End() {
                d.FieldStruct("instruction", func(d *decode.D) {
                        tag := d.FieldU8("tag", /* map 0 => "attestation", 0xff => "continuation" */)
                        // switch/if on tag
                })
        })
})
```
Then you can match on "attestation" instead of 0 byte a query `grep_by(.tag == "attestation") etc` (you can still access to raw "actual" value with `toactual` like `grep_by(.tag | toactual == 0)`)

About gaps: the sym mapping itself is not the thing that prevents a gap, just `d.FieldU8("attestation_tag")` is enough. Gap fields are added automatically for all bit ranges that no field is not "calming".

format/opentimestamps/ots.go
wader: Aha sorry it's in an array, if you use `dv` you will see the name. But maybe always having a "instruction" struct as i wrote about above it the way?

format/opentimestamps/ots.go
fiatjaf: I was going to do constants and mappings between constants and behavior and also check if an operation took an argument or not, since that's how the opentimestamps format is supposed be, a very flexible thing, extensible and so on -- but then I decided to write the simplest code possible that worked in practice and was easy to read.

I'll change to the abstracted approach.

format/opentimestamps/ots.go
fiatjaf: I had never seen such syntax and it doesn't seem to work on my `go`.

What I really wanted was to be able to match a sequence of 8 bytes against the 8 bytes magic sequences above, but I couldn't find a way. All I could find were the uint64 matchers. The BitBufs have confused me. What am I missing?

wader: Ah sorry i forgot to remove some "0x", so like this (the "_":s is just for clarity)
```go
	pendingMagic uint64 = 0x83_df_e3_0d_2e_f9_0c_8e
	bitcoinMagic uint64 = 0x05_88_96_0d_73_d7_19_01
```

Yes the BitBuf stuff should mostly be used when dealing with raw bytes or > 8 bytes things, and yeah i'm not very happy with the API for it either hmm

fiatjaf:
I think I got it all now.

I don't know what to put on an `ots.md` file. Where do these CLI docs show up?

I don't have a use case for filtering through these files actually, my initial reason for implementing this was to be able to "see" the entire file in one go, not so much to dig into its sections and so on.

wader:
> I think I got it all now.

👍 

> I don't know what to put on an `ots.md` file. Where do these CLI docs show up?

You can put a author section, links to references, examples and such, but it's optional.

You can show format help with `fq -h <format_name>` or in the REPL `help(<format_name>)`. It will also be used for `doc/formats.md` by `make doc`.

> I don't have a use case for filtering through these files actually, my initial reason for implementing this was to be able to "see" the entire file in one go, not so much to dig into its sections and so on.

Ok! but i think it's good to sym map numbers etc to human friendly things, not only for queries but also for when turning it into JSON.

format/opentimestamps/ots.go
```diff
@@ -0,0 +1,123 @@
...
+	keccak256Tag:     "keccak256",
+}
+
+func decodeOTSFile(d *decode.D) any {
+	d.Endian = decode.BigEndian
+
+	d.FieldRawLen("magic_bytes", int64(8*len(headerMagic)), d.AssertBitBuf(headerMagic))
+	d.FieldUintFn("version", decodeVarInt)
+	tag := d.FieldU8("digest_hash_algorithm", tagMapper)
+	if tag != 8 {
```
wader, line 68: Can be `tag == sha256Tag`?

fiatjaf: I forgot this one.

wader:
Ready to merge?

fiatjaf:
Should be ready now. Added the doc file with some simple examples.

wader:
Do one last `make doc` to add the docs to formats.md

wader:
Thanks!

fiatjaf:
Thank you very much for merging and for your patience!

wader:
No worries! also reminded me that i should look into minimizing number files needed to change to add a format

Issue #729, opened 2023-08-03
Color output is unreadable on terminals using light backgrounds.


adedomin:
### What version are you using (`fq -v`)?

<pre>
$ fq -v
0.7.0 (linux amd64)

</pre>

### How was fq installed?

<pre>nix shell 'nixpkgs#fq'</pre>

### Can you reproduce the problem using the latest release or master branch?

yes, 

### What did you do?

N/A

### What result did you expect?

colors safe for both dark bg and light bgs or no colors at all (-M flag is satisfactory for now)

### What did you see instead?

![Screenshot_20230802_212351](https://github.com/wader/fq/assets/6037590/46008127-ef91-4f3e-a23a-cd9b0bb485f2)

wader:
Hey! yes that does not look great. Do you have any other tool that uses color that work better? wonder if it's possible to come up with colors codes that work nice for both dark/bright background hmm.

I found this jq issue https://github.com/jqlang/jq/issues/495 that might give some inspiration.

If you want to help out and experiment you can do something like:
```
# see fq -h options for default
$ fq -o colors="array=white,dumpaddr=yellow,dumpheader=yellow+underline,error=brightred,false=yellow,index=white,null=brightblack,number=cyan,object=white,objectkey=brightblue,prompt_repl_level=brightblack,prompt_value=white,string=green,true=yellow,value=white" -o byte_colors="0-255=brightwhite,0=brightblack,32-126:9-13=white" ...
```

Color names are defined here https://github.com/wader/fq/blob/master/pkg/interp/ansi.jq and you can add more if needed.

wader:
What terminal and theme is this btw? trying to come up with a sane way to test things without to much effort. Something like this seems to work ok:

![Screenshot 2023-08-03 at 11 37 09](https://github.com/wader/fq/assets/185566/7fe9c5a0-80f6-47bc-b043-cf3b8354051b)

wader:
What about this?

![Screenshot 2023-10-10 at 10 56 59](https://github.com/wader/fq/assets/185566/5db8587b-730b-48ab-a609-fb8da6770e3b)

Generated with:
```
echo "Current"; fq . -o line_bytes=16 test.mp3 ; echo "New dark"; go run . -o line_bytes=16 -C . test.mp3; echo "New light"; go run . -o line_bytes=16 -C . test.mp3 | ansisvg -colorscheme "Builtin Solarized Light" | ffcat
```

wader:
Fixed by #781

PR #781, opened 2023-10-10
interp: Improve colors when using light background

wader:
Related to #729


wader:
@thushan hey! could you give this a spin in windows and see if colors looks ok in cmd and the new terminal thingy?

thushan:
Windows Terminal test (built from master 5af117474114751d62da1da56ce731d46a39a295)

**Dark**
image

**Light**
image

vs v0.8.0:

**Dark**
image

**Light**
image

This is probably an extreme example because I have Windows Terminal + Starship + dark themes, so the light theme isn't properly configured, but alas this does fix it :)

wader:
Thanks a lot! I should really get myself a windows dev env somehow, used to virtualized one but was slow pain to use.

Seems to look fine. A bit weird that neither before (brightwhite) or after (defaultgb+bold) the lightest white made any difference. Maybe a theme thing?

PR #824, opened 2023-12-04
leveldb: Add LevelDB support

mikez:


format/ldb/testdata/ldb.fqtest
```diff
@@ -0,0 +1,98 @@
+$ fq -d ldb dv uncompressed.ldb/000005.ldb
```
wader, line 1: snappy test will be added later?

format/ldb/testdata/make_ldb.py
```diff
@@ -0,0 +1,45 @@
+# Make LevelDB data: both uncompressed and compressed.
```
wader, line 1: Nice with code to generate test cases!

format/ldb/ldb.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/wader/fq/pkg/decode"
+	"github.com/wader/fq/pkg/interp"
+	"github.com/wader/fq/pkg/scalar"
+)
+
+func init() {
+	interp.RegisterFormat(
+		format.LDB,
+		&decode.Format{
+			Description: "LevelDB Table",
```
wader, line 25: You can add a .md file named the same as the format and add various tips, todos and author. see maybe bson.md and bson.go how it works. It will also be used for help in the CLI and for generating formats.md with `make doc`

wader (commented):
Nice start! did some initial reviewing to get going

format/format.go
```diff
@@ -125,6 +125,7 @@ var (
 	JPEG                = &decode.Group{Name: "jpeg"}
 	JSON                = &decode.Group{Name: "json"}
 	JSONL               = &decode.Group{Name: "jsonl"}
+	LDB                 = &decode.Group{Name: "ldb"}
```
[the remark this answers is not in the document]
mikez: Yes. And I'd go even one step further and use the `MPEG` pattern here of doing `leveldb_ldb` (and eventually `leveldb_log` etc.)

format/ldb/ldb.go
```diff
@@ -0,0 +1,279 @@
...
+		restartOffset = size*8 - (1+numRestarts)*uint32Size
+		d.SeekAbs(start + restartOffset)
+		d.FieldArray("restarts", func(d *decode.D) {
+			for i := 0; i < int(numRestarts); i++ {
+				d.FieldU32("restart")
+			}
+		})
+	})
+	// TK: how do you make an empty entries-array appear _above_ the trailer?
+	// Right now, its omited if empty.
```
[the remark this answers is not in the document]
mikez: Typically a key-value block has (1) entries and (2) trailer. However, occasionally there's only a trailer and no entries.
What to do in such scenarios?
Option 1: leave away the entries array all together. (Current solution.)
Option 2: Show an empty entries array. (Proposition. However, it ends up being shown below the trailer, since it's empty. We need to read the trailer first, to figure out if there's an entries array or not.)

format/ldb/ldb.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func readMetaContent(size int64, valueCallbackFn func(d *decode.D), d *decode.D) {
+	// TK(2023-12-04)
+	// https://github.com/google/leveldb/blob/main/doc/table_format.md#filter-meta-block
+	// https://github.com/google/leveldb/blob/main/table/filter_block.cc
+	d.FieldRawLen("raw", size*8)
+}
+
+func decodeVarInt(d *decode.D) uint64 {
```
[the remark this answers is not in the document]
mikez: Made a remark in tryULEB128 that it's the same as "Base 128 Varint" (a term used in Google contexts).

format/ldb/testdata/ldb.fqtest
mikez: added.

format/ldb/ldb.go
```diff
@@ -0,0 +1,279 @@
...
+	d.Endian = decode.LittleEndian
+
+	// footer
+
+	d.SeekAbs(d.Len() - footerEncodedLength*8)
+	var indexOffset int64
+	var indexSize int64
+	var metaIndexOffset int64
+	var metaIndexSize int64
+
```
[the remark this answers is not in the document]
mikez: Thanks for the pointer to these functions. Rewrote.

mikez:
@wader Also a question: what's your common way to hide "compressed" chunks when they're also decompressed. I added a `d.FieldRawLen("compressed",…)` to be consistent with other formats. However, when examining files, this can be distracting to have that compressed property show up as well. What's your default way to allow to hide the compressed parts if we're already showing the decompressed parts?

format/leveldb/leveldb_ldb.go
```diff
@@ -0,0 +1,265 @@
...
+			bb := &bytes.Buffer{}
+			switch compressionType {
+			case compressionTypeSnappy:
+				decompressed, err := snappy.Decode(nil, compressed)
+				if err != nil {
+					d.Fatalf("failed decompressing data: %v", err)
+				}
+				d.Copy(bb, bytes.NewReader(decompressed))
+			default:
+				d.Fatalf("Unsupported compression type: %x", compressionType)
```
wader, line 182: If it's possible to continue decoding even if there is an error you can use `d.Errorf` and then using `-f` will ignore the error

format/leveldb/leveldb_ldb.go
```diff
@@ -0,0 +1,265 @@
...
+			}
+			d.FieldStructRootBitBufFn("uncompressed", bitio.NewBitReader(bb.Bytes(), -1), func(d *decode.D) {
```
wader, line 184: I really should come up with a nicer API for the nested decoding stuff, feels a bit like a hack atm

pkg/decode/read.go
```diff
@@ -260,8 +265,8 @@ func (d *D) tryULEB128() (uint64, error) {
 		if shift >= 63 && b != 0 {
 			return 0, fmt.Errorf("overflow when reading unsigned leb128, shift %d >= 63", shift)
 		}
-		result |= (b & 0x7f) << shift
-		if b&0x80 == 0 {
+		result |= (b & 0b01111111) << shift
+		if b&0b10000000 == 0 {
```
wader, line 269: Nice clarifications here 👍

format/leveldb/leveldb_ldb.go
```diff
@@ -0,0 +1,265 @@
...
+
+func init() {
+	interp.RegisterFormat(
+		format.LDB,
+		&decode.Format{
+			Description: "LevelDB Table",
+			Groups:      []*decode.Group{format.Probe},
+			DecodeFn:    ldbDecode,
+		})
+	interp.RegisterFS(leveldbFS)
```
wader, line 31: Some formats has a `torepr` implementation which is a function to convert the decode tree inside more user representation, not sure if that is interesting for leveldb? for example bson has https://github.com/wader/fq/blob/master/format/bson/bson.jq also don't forget to register the format-overloaded--function https://github.com/wader/fq/blob/master/format/bson/bson.go#L24 (it's part for a hack to make is possible to have kind of "polymorphic" functions)

doc/formats.md
```diff
@@ -690,6 +691,23 @@ $ fq -n -d html '[inputs | {key: input_filename, value: .html.head.title?}] | fr
 $ fq -r -o array=true -d html '.. | select(.[0] == "a" and .[1].href)?.[1].href' file.html
 ```
 
+## leveldb_ldb
+
+### Limitations
+
+- no Meta Blocks (like "filter") are decoded yet.
+- Zstandard uncompression is not implemented yet.
```
wader, line 699: This is more or less just depend on some zstd package? i've looked at https://github.com/klauspost/compress a couple of times, it has zstd and lots of other formats and i also suspect the api:s are a bit more low level then golang stdlib so might fit fq better. It can maybe replace github.com/golang/snappy also? impressively it seems it also has zero dependencies on other packages

format/leveldb/testdata/ldb_uncompressed.fqtest
```diff
@@ -0,0 +1,98 @@
...
+     |                                               |                |    uncompressed{}: 0x60e-0x625 (23)
+     |                                               |                |      entries[0:1]: 0x60e-0x61d (15)
+     |                                               |                |        [0]{}: entry 0x60e-0x61d (15)
+0x600|                                          00   |              . |          shared_bytes: 0 0x60e-0x60f (1)
+0x600|                                             09|               .|          unshared_bytes: 9 0x60f-0x610 (1)
+0x610|03                                             |.               |          value_length: 3 0x610-0x611 (1)
+     |                                               |                |          key_delta{}: 0x611-0x61a (9)
+0x610|   73                                          | s              |            user_key: "s" 0x611-0x612 (1)
+0x610|      01                                       |  .             |            type: "value" (0x1) 0x612-0x613 (1)
+0x610|         ff ff ff ff ff ff ff                  |   .......      |            sequence_number: 72057594037927935 0x613-0x61a (7)
```
wader, line 78: Format as hex? but maybe decimal make sense if it's a sequence

format/format.go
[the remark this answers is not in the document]
wader: yeap makes sense, see my other comments about it

format/ldb/ldb.go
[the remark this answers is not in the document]
wader: yeap they are quite nice sometimes so one don't have to keep track number of bits left etc. not that i think `d.Pos()` inside a one of these still return the absolute position in the "root" buffer but `d.Len()` will change behaviour.  Maybe there should be `d.RootPos()`/`d.RootLen()` etc?

format/ldb/ldb.go
```diff
@@ -0,0 +1,279 @@
...
+		handle := BlockHandle{
+			Offset: d.FieldUintFn("offset", decodeVarInt),
+			Size:   d.FieldUintFn("size", decodeVarInt),
+		}
+		metaHandles = append(metaHandles, handle)
+	}, d)
+
+	// index
+
+	d.SeekAbs(indexOffset * 8)
```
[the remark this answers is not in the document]
wader: No worries, only use if it feels it makes things clearer. `d.RangeFn` can sometimes be used instead of combination of seek and the limit

format/ldb/ldb.go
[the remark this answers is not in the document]
wader: I guess it would be nice to have an empty entries array for clarity? i will have to look closer at how to solve this. But don't think it's a blocker to merge, can be changed/fixed later.

wader:
> @wader Also a question: what's your common way to hide "compressed" chunks when they're also decompressed?
> 
> Background: I added a `d.FieldRawLen("compressed",…)` to be consistent with other formats. However, when examining files, this can be distracting to have that compressed property show up as well. What's your default way to allow to hide the compressed parts if we're already showing the decompressed parts?

I usually lean towards not hide things as that is kind of what fq is all about. In this case if you did not add "compressed" fields would those bit ranges show as gaps fields instead?

wader:
The failing test is about `TestFormats/all/all.fqtest`, probably just a matter of `WRITE_ACTUAL=1 go test -run TestFormats/all/all.fqtest ./format`. I've thought about redoing those tests... would be nice if adding a format would affect as little as possible outside the formats own directory

format/format.go
```diff
@@ -125,6 +125,7 @@ var (
 	JPEG                = &decode.Group{Name: "jpeg"}
 	JSON                = &decode.Group{Name: "json"}
 	JSONL               = &decode.Group{Name: "jsonl"}
+	LDB                 = &decode.Group{Name: "leveldb_ldb"}
```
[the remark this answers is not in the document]
mikez: @wader 👌, good idea. Besides the suggested modification, I now also added `leveldb_log` and `leveldb_descriptor`; thereby keeping the names as in the LevelDB [`leveldbutils dump` command](https://github.com/google/leveldb/blob/main/db/dumpfile.cc#L214).

format/leveldb/leveldb_ldb.go
```diff
@@ -0,0 +1,265 @@
...
+	compressionTypeZstandard: "zstd",
+}
+
+// https://github.com/google/leveldb/blob/main/db/dbformat.h#L54
+var valueTypes = scalar.UintMapSymStr{
+	0x0: "deletion",
+	0x1: "value",
+}
+
+type BlockHandle struct {
```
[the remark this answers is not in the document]
mikez: Ah, privacy in Go is solved by title-casing! Interesting language design decision.

format/leveldb/leveldb_ldb.go
mikez: 👍 Replaced all d.Fatalf with d.Errorf where it made sense.

format/leveldb/leveldb_ldb.go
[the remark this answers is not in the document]
wader: yeap, let's say you learn to be ok with it :) not sure how well it agrees with golang's principle of clarity and explicitness 🤔

format/leveldb/leveldb_ldb.go
mikez: Thanks for the pointers. It's not clear to me in which scenarios you'd use `torepr`.

format/leveldb/testdata/ldb_uncompressed.fqtest
mikez: Typically sequence makes sense in decimal. They are sequential numbers. The only exception are these index  and metaindex sections; especially the example you highlighted, where the maximum possible unsigned integer `0xffffffffffffff` is chosen.

mikez:
> > @wader Also a question: what's your common way to hide "compressed" chunks when they're also decompressed?
> > Background: I added a `d.FieldRawLen("compressed",…)` to be consistent with other formats. However, when examining files, this can be distracting to have that compressed property show up as well. What's your default way to allow to hide the compressed parts if we're already showing the decompressed parts?
> 
> I usually lean towards not hide things as that is kind of what fq is all about. In this case if you did not add "compressed" fields would those bit ranges show as gaps fields instead?

I'd like you to hear me differently.

Say there is a chunk of data that is compressed with Snappy. I uncompress this data and show the uncompressed data-structure under the key "uncompressed". However, in line with other formats, I also include a key called "compressed" which has the original raw Snappy-compressed bits. When there are a lot of compressed sections like these (which have all been uncompressed and have a corresponding "uncompressed" key) it can get quite unwieldy to skim the output. Therefore the inquiry into if it makes sense to hide these "compressed" sections (or more forcibly truncating them in the preview) if there are already corresponding "uncompressed" sections.

format/leveldb/testdata/ldb_uncompressed.fqtest
wader: Ok! keep as decimal. btw if 0xffffffffffffff has some special mening you can add description. For example the mp4 decoder does something like this for box size:
```go
const (
	boxSizeRestOfFile   = 0
	boxSizeUse64bitSize = 1
)

d.FieldU32("size", scalar.UintMapDescription{
	boxSizeRestOfFile:   "Rest of file",
	boxSizeUse64bitSize: "Use 64 bit size",
})
```
In your case you could probably just have `d.FieldU64("...", scalar.UintMapDescription{0xffffffffffffff: "blabla"})`

format/leveldb/leveldb_ldb.go
mikez: @wader Ah, I think I get it! This was a helpful example for me. Thank you. :)
Yes, this applies here, especially for descriptors. However, this reminds me of another question I have. I will ask below.

format/leveldb/testdata/ldb_uncompressed.fqtest
mikez: @wader It's unclear to me if it has a special meaning. I only noticed so far that in the index and metaindex there are sometimes sequence numbers like that; it feels more like a hack though. The sequence numbers are typically used in the data section and there they have a well-defined meaning.

wader:
> I'd like you to hear me differently.
> 
> Say there is a chunk of data that is compressed with Snappy. I uncompress this data and show the uncompressed data-structure under the key "uncompressed". However, in line with other formats, I also include a key called "compressed" which has the original raw Snappy-compressed bits. When there are a lot of compressed sections like these (which have all been uncompressed and have a corresponding "uncompressed" key) it can get quite unwieldy to skim the output. Therefore the inquiry into if it makes sense to hide these "compressed" sections (or more forcibly truncating them in the preview) if there are already corresponding "uncompressed" sections.

I see. As it works currently if you didn't add "compressed" fields they would instead show up as "gap" fields that fq automatically inserts, this is so that all bits will always be "reachable" somehow. Maybe what you looking for is the ability to add a "compressed" but somehow tell fq that it should be displayed in a more discreet way unless is verbose mode etc? I'm thinking if totally hiding a field might be confusing as it will look as there is a gap in data/hex column? or maybe i misunderstand what your aiming for?

format/leveldb/testdata/ldb_uncompressed.fqtest
wader: Ok! then sounds like we should leave as it is

mikez:
@wader Question about the LevelDB log format and how to decompress it.

In essence the data structure is as follows:
1. We have a sequence of records, each with some data. The user can iterate over these using the LevelDB API.
2. In the `.log`-file itself, these records are split into many small pieces and put into blocks of 32KB. Each little piece has a marker if it's an entire record ("full") or only a fragment ("first", "middle", "last").
3. In the LevelDB app itself, these small little pieces are put together by and the datastructure is parsed.

I hope that's clear so far. 

Thus my question is: for the pieces which are preserved in full, it's easy to show its underlying datastructure, since it hasn't been split. However, for the "first", "middle", and "last" pieces, I wouldn't know how to visualize it? Is there some precedence here in the other formats?

format/format.go
```diff
@@ -125,6 +125,9 @@ var (
 	JPEG                = &decode.Group{Name: "jpeg"}
 	JSON                = &decode.Group{Name: "json"}
 	JSONL               = &decode.Group{Name: "jsonl"}
+	LevelDB_Descriptor  = &decode.Group{Name: "leveldb_descriptor"}
+	LDB                 = &decode.Group{Name: "leveldb_table"}
+	LOG                 = &decode.Group{Name: "leveldb_log"}
```
wader, line 130: Prefix with `LevelDB_`

format/leveldb/leveldb_descriptor.go
```diff
@@ -0,0 +1,125 @@
...
+	return nil
+}
+
+// List of sorted tables for each level involving key ranges and other metadata.
+func readManifest(d *decode.D) {
+	d.FieldArray("tags", func(d *decode.D) {
+		for {
+			if d.End() {
+				break
+			}
```
wader, line 76: Could this be `for !d.End() {`?

mikez:
@wader Something about those test failures I don't quite understand; it seems some aspect of LevelDB is leaking into the other tests involving `FieldFormatBitBuf`.

format/leveldb/leveldb_log.go
```diff
@@ -0,0 +1,152 @@
...
+
+//go:embed leveldb_log.md
+var leveldbLogFS embed.FS
+
+func init() {
+	interp.RegisterFormat(
+		format.LOG,
+		&decode.Format{
+			Description: "LevelDB Log",
+			Groups:      []*decode.Group{format.Probe},
```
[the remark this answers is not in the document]
mikez: I was not quite aware what that `format.Probe` was doing (I copy pasted that part); now I am. Since the log files don't have magic strings of any kind in them, I think it's better to exclude them for now + from descriptors which is based on similar logic.

wader:
> @wader Question about the LevelDB log format and how to decompress it.
> 
> In essence the data structure is as follows:
> 
> 1. We have a sequence of records, each with some data. The user can iterate over these using the LevelDB API.
> 2. In the `.log`-file itself, these records are split into many small pieces and put into blocks of 32KB. Each little piece has a marker if it's an entire record ("full") or only a fragment ("first", "middle", "last").
> 3. In the LevelDB app itself, these small little pieces are put together by and the datastructure is parsed.
> 
> I hope that's clear so far.
> 
> Thus my question is: for the pieces which are preserved in full, it's easy to show its underlying datastructure, since it hasn't been split. However, for the "first", "middle", and "last" pieces, I wouldn't know how to visualize it? Is there some precedence here in the other formats?

Think i would try to be "true" to how the format works, so show the parts as some array with structs with raw data field for the partia data etc. This is how most formats in fq work that has to "demux" etc, ex ogg, gzip and the tcp reassembly. This also makes it quite nice to when deal with broken files that might have some partial data that is good, then you can use fq to concat parts, maybe prefix/append some missing data and decode or output the "repaired" data.

format/leveldb/leveldb_log.go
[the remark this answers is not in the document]
wader: 👍 yeap there is no special probe code in fq, just that some decoders are tried if none is specified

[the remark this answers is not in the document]
wader: extra trivia: all formats and group exist as jq functions so you can do this ex: `... | probe`

mikez:
> > I'd like you to hear me differently.
> > Say there is a chunk of data that is compressed with Snappy. I uncompress this data and show the uncompressed data-structure under the key "uncompressed". However, in line with other formats, I also include a key called "compressed" which has the original raw Snappy-compressed bits. When there are a lot of compressed sections like these (which have all been uncompressed and have a corresponding "uncompressed" key) it can get quite unwieldy to skim the output. Therefore the inquiry into if it makes sense to hide these "compressed" sections (or more forcibly truncating them in the preview) if there are already corresponding "uncompressed" sections.
> 
> I see. As it works currently if you didn't add "compressed" fields they would instead show up as "gap" fields that fq automatically inserts, this is so that all bits will always be "reachable" somehow. Maybe what you looking for is the ability to add a "compressed" but somehow tell fq that it should be displayed in a more discreet way unless is verbose mode etc? I'm thinking if totally hiding a field might be confusing as it will look as there is a gap in data/hex column? or maybe i misunderstand what your aiming for?

Yes, I hear you. I think I found a solution... to use `d` instead of `dd`. That nicely truncates the compressed parts into one line. Some of the regular strings might be truncated as well like that, but I could look them up manually as needed.

format/leveldb/leveldb_table.go
```diff
@@ -82,9 +82,13 @@ func ldbTableDecode(d *decode.D) any {
 	var metaIndexOffset int64
 	var metaIndexSize int64
 
-	d.SeekAbs(d.Len() - footerEncodedLength)
 	d.FieldStruct("footer", func(d *decode.D) {
-		handleLength := d.LimitedFn(footerEncodedLength, func(d *decode.D) {
+		// check for magic number and fail fast if it isn't there
+		d.SeekAbs(d.Len() - magicNumberLength)
+		d.FieldU64("magic_number", d.UintAssert(tableMagicNumber), scalar.UintHex)
```
wader, line 88: If you want you can probably keep it like it was, as long as it does not decode megabytes of data before failing it should probably be ok.

wader:
> Yes, I hear you. I think I found a solution... to use `d` instead of `dd`. That nicely truncates the compressed parts into one line. Some of the regular strings might be truncated as well like that, but I could look them up manually as needed.

Aha that explains things :) i wonder if we could have separate options for raw and string truncate limit? by look up manually you mean use a query to access a string field?

mikez:
> > Yes, I hear you. I think I found a solution... to use `d` instead of `dd`. That nicely truncates the compressed parts into one line. Some of the regular strings might be truncated as well like that, but I could look them up manually as needed.
> 
> Aha that explains things :) i wonder if we could have separate options for raw and string truncate limit? by look up manually you mean use a query to access a string field?

use a query, exactly! In my imaginary world, the entire thing would be an App like Hex Fiend or some debugger, and I could just hover over it to see the full thing as a tooltip.

doc/formats.md
wader: Ok, then i think we can skip it for now, probably good to not make this PR grow too much also

wader:
> use a query, exactly! In my imaginary world, the entire thing would be an App like Hex Fiend or some debugger, and I could just hover over it to see the full thing as a tooltip.

Yeap that would very interesting. I can imagine some kind of "IDE" with multiple visual tree:s, data views and REPL:s.

format/leveldb/leveldb_descriptor.go
```diff
@@ -0,0 +1,121 @@
...
+	tagTypeDeletedFile    = 6
+	tagTypeNewFile        = 7
+	// 8 not used anymore
+	tagTypePrevLogNumber = 9
+)
+
+var tagTypes = scalar.UintMapSymStr{
+	tagTypeComparator:     "comparator",
+	tagTypeLogNumber:      "log_number",
+	tagTypeNextFileNumber: "next file number",
```
wader, line 46: Use underscore for these? i usually try use jq-ish constant unless the specification is very explicit about the naming

README.md
```diff
@@ -102,6 +102,7 @@ ipv6_packet,
 jpeg,
 json,
 jsonl,
+[leveldb_ldb](doc/formats.md#leveldb_ldb),
```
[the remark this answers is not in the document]
mikez: Oddly, the `make doc` commands changes a lot of content in the SVGs. I left them out in the commit, as it seems they shouldn't change.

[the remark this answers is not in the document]
wader: Aha yes skip those, i will fix. I did some changes to https://github.com/wader/ansisvg some days ago

format/leveldb/leveldb_descriptor.jq
```diff
@@ -0,0 +1,42 @@
...
+          ) else . end
+        | if .compact_pointer then (
+            .compact_pointer.internal_key |= .data
+          ) else . end
+      )
+    end
+   );
+  ( {type: "root", value: .}
+  | _f
+  );
```
wader, line 42: To keep the jq style same in fq use some kind of "haskell"-ish inspired style, something like:
```jq
def _leveldb_descriptor_torepr:
  def _f:
    if .type == "root" then
      [ .value.blocks[].records[]
      | {type: "record", value: .}
      | _f
      ]
    elif .type == "record" then
      if .value.header.record_type != "full" then
        empty
      else
        [ .value.data.tags[]
        | {type: "tag", value: {(.key): .value}}
        | _f
        ]
      end
    elif .type == "tag" then
      ( .value
      | if .comparator then .comparator |= .data else . end
      | if .new_file then
          ( .new_file.largest_internal_key |= .data
          | .new_file.smallest_internal_key |= .data
          )
        else .
        end
      | if .compact_pointer then
          .compact_pointer.internal_key |= .data
        else .
        end
      )
    end;
  ( {type: "root", value: .}
  | _f
  );

```

format/leveldb/leveldb_table.go
```diff
@@ -0,0 +1,281 @@
...
+
+func init() {
+	interp.RegisterFormat(
+		format.LevelDB_LDB,
+		&decode.Format{
+			Description: "LevelDB Table",
+			Groups:      []*decode.Group{format.Probe},
+			DecodeFn:    ldbTableDecode,
+		})
+	interp.RegisterFS(leveldbFS)
```
wader, line 35: For consistency `leveldbTableFS`?

format/leveldb/leveldb_table.go
```diff
@@ -0,0 +1,281 @@
...
+		d.SeekAbs(start)
+		if compressionType == compressionTypeNone {
+			d.FieldStruct("uncompressed", func(d *decode.D) {
+				readTableBlockContents(size, valueCallbackFn, d)
+			})
+		} else {
+			compressedSize := size
+			compressed := data
+			bb := &bytes.Buffer{}
+			fmt.Println(bb, bb.Len())
```
wader, line 183: Debug leftover i guess

wader:
Played around a bit with the leveldb table decoder locally and noticed this, looks a bit fishy?
```sh
➜  fq git:(leveldb) ✗ go run . -o line_bytes=8 -d leveldb_table '.metaindex | d' ~/Library/Application\ Support/Spotify/PersistentCache/public.ldb/000293.ldb
2023/12/07 15:04:17 bla1
2023/12/07 15:04:17 bla1
2023/12/07 15:04:17 bla1
      │00 01 02 03 04 05 06 07│01234567│.metaindex{}:
      │                       │        │  uncompressed{}:
      │                       │        │    entries[0:1]:
      │                       │        │      [0]{}: entry
0x7a58│         00            │   .    │        shared_bytes: 0
0x7a58│            22         │    "   │        unshared_bytes: 34
0x7a58│               05      │     .  │        value_length: 5
      │                       │        │        internal_key{}:
0x7a58│                  66 69│      fi│          user_key: "filter.leveldb.BuiltinBloo"
0x7a60│6c 74 65 72 2e 6c 65 76│lter.lev│
0x7a68│65 6c 64 62 2e 42 75 69│eldb.Bui│
0x7a70│6c 74 69 6e 42 6c 6f 6f│ltinBloo│
0x7a78│6d                     │m       │          type: 0x6d
0x7a78│   46 69 6c 74 65 72 32│ Filter2│          sequence_number: 14199528906058054
      │                       │        │        value{}:
0x7a80│97 f2 01               │...     │          offset: 30999
0x7a80│         bf 02         │   ..   │          size: 319
      │                       │        │    trailer{}:
      │                       │        │      restarts[0:1]:
0x7a80│               00 00 00│     ...│        [0]: 0
0x7a88│00                     │.       │
0x7a88│   01 00 00 00         │ ....   │      num_restarts: 1
0x7a88│               00      │     .  │  compression: "none" (0x0)
0x7a88│                  71 a3│      q.│  checksum: 0x51caa371 (valid)
0x7a90│ca 51                  │.Q      │
```

Also noticed some strings seems to be length prefixed `user_key: "\x11PlaylistSyncEvent#"` but maybe that is just how the application saves it's data?

wader:
> Great find! Two things:
> 
> 1. I need to check out the metaindex again. I falsely supposed it followed the same format as index, but it clearly doesn't.
> 2. Yes, to "save space" common prefixes are truncated. The "unshared" and "shared" properties keep track of the bytes that may need to get prefixed to keys to form the entire key. In the view, I show what is actually stored in the database, not what the app would eventually generate.

I see, if you want and if it's handy for a user you can add "synthetic" fields that have the "unshares" strings somehow `d.FieldValue<type>("...", ...)`. Synthetic fields are fields that are not backed by any bitrange so they are just jq values. Can be quite handy for extra annotation or maybe derived values that are common for the format.

mikez:
> d.FieldValue

I tried to search for "synthetic field" in the code, but couldn't find any examples, except for the flag in scalar.go.

wader:
> > d.FieldValue
> 
> I tried to search for "synthetic field" in the code, but couldn't find any examples, except for the flag in scalar.go. How would I implement this?

Hmm strange, but it was quite recently that i made synthetic something more special, before they were a hack using zero length bit ranges so it might explain why "synthetic" is not used much in the code yet.

For example mp4 decoder adds track id and format here https://github.com/wader/fq/blob/master/format/mp4/mp4.go#L280-L291

mikez:
> Stress tested a bit more using:
> 
> ```
> find ~/ -name "*.ldb" -print0 | xargs -0 go run . -d leveldb_table '._error | select(.) | input_filename, .error'
> ```
> 
> Seems chrome has a bunch of leveldb files, some fail with variations of `"UTF8(user_key): failed at position 127 (read size 0 seek pos 0): tryText nBytes must be >= 0 (-1)`

I appreciate the stress-testing and the command. I'll take a look tomorrow!

wader:
> I appreciate the stress-testing and the command. I'll take a look tomorrow!

👍 no stress! i think the PR is in very good shape now, some more testing and fix decode issues and we should be ready to merge. anything else you want to add?

mikez:
@wader That one took me a while to track down and fix. :)

mikez:
> > > d.FieldValue
> > 
> > 
> > I tried to search for "synthetic field" in the code, but couldn't find any examples, except for the flag in scalar.go. How would I implement this?
> 
> Hmm strange, but it was quite recently that i made synthetic something more special, before they were a hack using zero length bit ranges so it might explain why "synthetic" is not used much in the code yet.
> 
> For example mp4 decoder adds track id and format here https://github.com/wader/fq/blob/master/format/mp4/mp4.go#L280-L291

Thank you. I searched for "synthetic"; had I searched for "FieldValue", then I would have found it faster. I wrongly assumed FieldValue was just a regular reader. :)

format/leveldb/leveldb_descriptor.jq
mikez: @wader Maybe you could get inspired by https://github.com/mvdan/sh ?

format/leveldb/leveldb_descriptor.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func readTagInternalKey(name string, d *decode.D) {
+	d.FieldStruct(name, func(d *decode.D) {
+		length := d.FieldULEB128("length")
+		d.FieldStruct("data", func(d *decode.D) {
+			err := readInternalKey(nil, int(length), d)
+			if err != nil {
+				// TK(2023-12-08): how do I propagate this
+				// error `err` into the `d` object?
```
wader, line 125: There is `d.IOPanic` that i think is similar? maybe there should be a similar one for generic error, or maybe `d.Errorf` should support `%w` error arguments somehow? have to read up on how that works. For now i think it can be fine to do `d.Errof("blabla: %s", err)` and format out the error as a string maybe?

format/leveldb/testdata/leveldb_table_uncompressed.fqtest
```diff
@@ -0,0 +1,100 @@
...
+0x010|                     4c 6f 72 65 6d 20 69 70 73|       Lorem ips|            value: "Lorem ipsum dolor sit amet, consectetur adipisc..." 0x17-0x1d4 (445)
+0x020|75 6d 20 64 6f 6c 6f 72 20 73 69 74 20 61 6d 65|um dolor sit ame|
+*    |until 0x1d3.7 (445)                            |                |
+     |                                               |                |          [1]{}: entry 0x1d4-0x3a2 (462)
+0x1d0|            06                                 |    .           |            shared_bytes: 6 0x1d4-0x1d5 (1)
+0x1d0|               0d                              |     .          |            unshared_bytes: 13 0x1d5-0x1d6 (1)
+0x1d0|                  bd 03                        |      ..        |            value_length: 445 0x1d6-0x1d8 (2)
+     |                                               |                |            key{}: 0x1d8-0x1e5 (13)
+0x1d0|                        69 70 73 75 6d         |        ipsum   |              user_key_suffix: raw bits 0x1d8-0x1dd (5)
+     |                                               |                |              user_key: "lorem.ipsum" (inferred)
```
wader, line 24: Nice :)

format/leveldb/testdata/leveldb_table_uncompressed.fqtest
```diff
@@ -0,0 +1,100 @@
...
+0x010|03 00 00 00 00 00 00                           |.......         |              sequence_number: 3 0x10-0x17 (7)
+0x010|                     4c 6f 72 65 6d 20 69 70 73|       Lorem ips|            value: "Lorem ipsum dolor sit amet, consectetur adipisc..." 0x17-0x1d4 (445)
+0x020|75 6d 20 64 6f 6c 6f 72 20 73 69 74 20 61 6d 65|um dolor sit ame|
+*    |until 0x1d3.7 (445)                            |                |
+     |                                               |                |          [1]{}: entry 0x1d4-0x3a2 (462)
+0x1d0|            06                                 |    .           |            shared_bytes: 6 0x1d4-0x1d5 (1)
+0x1d0|               0d                              |     .          |            unshared_bytes: 13 0x1d5-0x1d6 (1)
+0x1d0|                  bd 03                        |      ..        |            value_length: 445 0x1d6-0x1d8 (2)
+     |                                               |                |            key{}: 0x1d8-0x1e5 (13)
+0x1d0|                        69 70 73 75 6d         |        ipsum   |              user_key_suffix: raw bits 0x1d8-0x1dd (5)
```
wader, line 23: There are not strings?

wader:
> Thank you. I searched for "synthetic"; had I searched for "FieldValue", then I would have found it faster. I wrongly assumed FieldValue was just a regular reader. :)

Aha sorry should have been clearer :)

BTW now all *.ldb files in my home directory decodes fine and looks beautiful!

format/leveldb/testdata/leveldb_table_uncompressed.fqtest
mikez: @wader They _are_; I couldn't figure out if you had a scalar method I could plug into `d.FieldRawLen`... since I need the bytes as well:

```
br := d.FieldRawLen("user_key_suffix", int64(unsharedSize-typeAndSequenceNumberSize)*8)
```

format/leveldb/testdata/leveldb_table_uncompressed.fqtest
mikez: @wader This decoding has given me a new insight into everything LevelDB stores; not just the key and the value, but apparently also the history of key-value pairs.

format/leveldb/leveldb_descriptor.go
mikez: `err.Error()` seems to do the trick.

format/leveldb/leveldb_table.go
```diff
@@ -0,0 +1,443 @@
...
+		d.FieldUTF8("user_key", keySize-typeAndSequenceNumberSize)
+		d.FieldU8("type", valueTypes, scalar.UintHex)
+		d.FieldU56("sequence_number")
+		return nil
+	}
+
+	// case 2: type and sequence_number fit fully in unshared: simulate user_key value.
+	if unsharedSize >= typeAndSequenceNumberSize {
+		br := d.FieldRawLen("user_key_suffix", int64(unsharedSize-typeAndSequenceNumberSize)*8)
+		d.FieldValueStr("user_key", string(append(sharedBytes, d.ReadAllBits(br)...)), strInferred)
```
wader, line 364: Could it be something like?
```go
suffix := d.FieldUTF8("user_key_suffix", int64(unsharedSize-typeAndSequenceNumberSize))
d.FieldValueStr("user_key", string(sharedBytes)+suffix), strInferred)
```

An alternativ to using `d.FieldValueStr` could be to only have one `user_key` field and let the symbolic value for that one string be the inferred string. But i think i like how it is now which i guess is more true to how thing actually work, that it's a suffix, then and just add a bit extra to make it convenient.

btw `append(sharedBytes, d.ReadAllBits(br)...)` can something in go be a bit tricky, if i remember correctly the `sharedBytes` slice might end up being reused if it as capacity, but in this case it should be fine i think as it get turned into a new string? (which immutable and not share the slice bytes)

format/leveldb/testdata/leveldb_table_uncompressed.fqtest
wader: 🥳  yeap i've had similar experiences writing decoders, you bump into lots of fascinating tricks and legacy, ex i remember zip has so much legacy to handle floppy fdisks and appending and also the date format uses msdos-timestamps (2 second precision!)

format/leveldb/testdata/leveldb_table_uncompressed.fqtest
wader: Aha ok, the suffix bytes are not always valid utf8? anyways i think the current way is fine! but i do wonder if there should be a `Raw` variant that return bytes? (maybe there is?) the reason `Raw` returns a `bitio.BitReader` is that they can be used for very large bit ranges so that the bits will only be read if really needed to save on IO and memory.

format/leveldb/leveldb_table.go
```diff
@@ -0,0 +1,443 @@
...
+				if keyCallbackFn == nil && shared == 0 {
+					d.FieldUTF8("key", int(unshared))
+				} else {
+					d.FieldStruct("key", func(d *decode.D) {
+						if keyCallbackFn == nil {
+							keyCallbackFn = readPrefixedBytes
+						}
+						err := keyCallbackFn(keyPrefix, int(unshared), d)
+						if err != nil {
+							d.Errorf(err.Error())
```
wader, line 280: Maybe these should be `d.Errorf("%s", err)` to be sure some random `%s` etc in the error string is seen as a format?

format/leveldb/leveldb_descriptor.go
```diff
@@ -0,0 +1,128 @@
...
+	})
+}
+
+func readTagInternalKey(name string, d *decode.D) {
+	d.FieldStruct(name, func(d *decode.D) {
+		length := d.FieldULEB128("length")
+		d.FieldStruct("data", func(d *decode.D) {
+			err := readInternalKey(nil, int(length), d)
+			if err != nil {
+				d.Errorf(err.Error())
```
wader, line 124: See other comment about `d.Errorf(err.Error())`

format/leveldb/leveldb_table.go
mikez: @wader I like your solution.
"the sharedBytes slice might end up being reused"
the `append()` of Go I don't quite understand yet. Apparently it is an amoritized-O(1) operation yet it seems sharedBytes doesn't get mutated? (So it's not doing `.append(....)` like in Python or `.push` like in JavaScript.)

wader:
Great work! just some tiny things left i think, after that just let me know if you think your read to merge

Hope it was a good review experience too 😄

format/leveldb/leveldb_table.go
mikez: @wader Although, on second thought, if the user_key is cut at some UTF-8 multi-byte character, this might fail. However, maybe it is good enough.

format/leveldb/leveldb_table.go
mikez: Thank you for the pointer. I just saw earlier that I'd used some `d.Errorf("%v", err)` code I had copy-pasted from avro_ocf.go without remembering.

format/leveldb/leveldb_table.go
mikez: In testing, I see it does indeed fail:

image

mikez: In LevelDB, both keys and values are bytestrings; so a UTF-8 encoding would typically be used, but most not necessarily.

mikez:
> Great work! just some tiny things left i think, after that just let me know if you think your read to merge
> 
> Hope it was a good review experience too 😄

Thank _you_, @wader! I very much appreciate your reviewing style. I learned a lot about fq, Go, and LevelDB in the process. I hope it was a good review experience for you as well.

format/leveldb/leveldb_table.go
mikez: @mikez I made some own functions for now to handle this. See recent commit.

format/leveldb/leveldb_table.go
```diff
@@ -441,3 +446,21 @@ func mask(crc uint32) uint32 {
 	// Rotate right by 15 bits and add a constant.
 	return ((crc >> 15) | (crc << 17)) + kMaskDelta
 }
+
+// Concatinate byteslices and convert into a string.
+func stringify(byteSlices ...[]byte) string {
```
wader, line 451: If you want to save on allocations you could use a `strings.Builder` or possible sum all lengths allocate once with `make()` and then append. But i have feeling it wont make much difference in this case?

format/leveldb/leveldb_table.go
```diff
@@ -441,3 +446,21 @@ func mask(crc uint32) uint32 {
...
+	var result []byte
+	for _, b := range byteSlices {
+		result = append(result, b...)
+	}
+	return string(result)
+}
+
+func fieldUTF8ReturnBytes(name string, nBytes int, d *decode.D) []byte {
```
wader, line 459: I've experiment and thought a bit about doing a decode api v2 that would use method chaining then things could probably be nicer, something like:
```
d.FieldUTF8("bla", nBytes).Bytes()
d.FieldUTF8("bla", nBytes).BitReader()
d.FieldUTF8("bla", nBytes).Map(...).Bytes()
```
Think that would make it a bit more flexible, would probably also cut down a bit on amount of generated code. Also! it can probably help make the api more type safe as ex `d.FieldUTF8` would return some type forcing things to use string. Should think more about this :)

format/leveldb/leveldb_table.go
mikez: image

`strings.Builder` seems to do the same thing. I suppose the resizing would be made in an amortized-constant way (similar to hash tables)? I can still preallocate though. :)

mikez: image

Ok, this explains more what's going on. It seems sometimes it's mutated (and then I assume that's amortized constant)... and sometimes it's created from scratch (which is costly). It is not clear to me which one happens when though.

format/leveldb/leveldb_table.go
```diff
@@ -360,8 +365,8 @@ func readInternalKey(sharedBytes []byte, unsharedSize int, d *decode.D) error {
 
 	// case 2: type and sequence_number fit fully in unshared: simulate user_key value.
 	if unsharedSize >= typeAndSequenceNumberSize {
-		br := d.FieldRawLen("user_key_suffix", int64(unsharedSize-typeAndSequenceNumberSize)*8)
-		d.FieldValueStr("user_key", string(append(sharedBytes, d.ReadAllBits(br)...)), strInferred)
+		suffix := fieldUTF8ReturnBytes("user_key_suffix", unsharedSize-typeAndSequenceNumberSize, d)
+		d.FieldValueStr("user_key", stringify(sharedBytes, suffix), strInferred)
```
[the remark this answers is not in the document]
mikez: Ah, interesting point with UTF8BOM (0xEF, 0xBB, 0xBF)!

I hadn't come across this before nor thought about it.
It seems the Python encoder natively skips it (unless specified explicitly), also some of the LevelDB dumps I've observed so far don't use the UTF8BOM.
That said, since LevelDB takes bytestrings as keys and values, that wouldn't stop anyone from using UTF8BOM if they wanted to.

So maybe the utf-8 encoder will be good enough in most cases?

format/leveldb/leveldb_table.go
wader: Huh i assumed `strings.Builder` did more smart things by growing smarty but maybe the only smartness is in the `String()` method? but i do wonder if append might have some smartness about how to grow so it won't end up reallocating/copy all the time?

format/leveldb/leveldb_table.go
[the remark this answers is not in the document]
wader: Safest i think is probably to keep it as is, treat the parts as raw bytes. It's a bit weird that UTF8 with a BOM is even a thing as UTF8 does not really have a byte order :) but it seems to happen 🤷 the encoding used by `d.UTF8` is choose to be quite liberal with weird stuff :)

BTW if you want to have full control over text encoding then there is `d.FieldStr("...", <encoding>)` where you can plugin any `encoding.Encoding`.

wader:
> > Great work! just some tiny things left i think, after that just let me know if you think your read to merge
> > Hope it was a good review experience too 😄
> 
> Thank _you_, @wader! I very much appreciate your reviewing style. I learned a lot about fq, Go, and LevelDB in the process. I hope it was a good review experience for you as well.

Happy to hear that! i can sometimes end up thinking that i might come a cross as "naggy" when there is a lot of back and forth, i'm not! superhappy someone wants to help out :) also a bit tricky to review things about the format itself as i have very little experience with it, trying to mostly focus on helping it fit well into rest of fq :)

format/leveldb/testdata/lorem.json
```diff
@@ -0,0 +1,6 @@
...
+    "lorem.ipsum": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.",
+    "lorem.dolor": "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.",
+    "row": "Row, row, row your boat\nGently down the stream.\nMerrily, merrily, merrily, merrily,\nLife is but a dream. 🚣‍♂️"
```
wader, line 5: Now i noticed the 🚣‍♂️ :)

wader:
Played around a bit more with chrome leveldb:s, seems to work fine! is a bit confusing to look at web storage databases, lots of weird stuff stored in them :)

Think i'm kind of ready to merge if you are

format/leveldb/leveldb_table.go
mikez: @wader I'm trying to track down how append works precisely; this is closest I've found: https://github.com/golang/go/blob/master/src/cmd/compile/internal/ssagen/ssa.go#L3503
Don't have more clarity now. :)

mikez:
@wader I'm ready! :)

format/leveldb/leveldb_table.go
wader: interesting! it seems to generate a call to growslice https://github.com/golang/go/blob/fe1b2f95e6dbfb6e6212bb391706ae62eb0ae5ec/src/runtime/slice.go#L155 which in turn seems to call nextslicecaphttps://github.com/golang/go/blob/fe1b2f95e6dbfb6e6212bb391706ae62eb0ae5ec/src/runtime/slice.go#L267 which has some heuristics how to grow

wader:
🥳

format/leveldb/leveldb_table.go
wader: https://godbolt.org/z/rEPerb3b4 maybe interesting. i would have assumed that append would just end up being a call to something but seem there lots of tricky going on, maybe be speed up common cases etc?

mikez: @wader Oh, checking the compiler explorer was very insightful!! `runtime.growslice(SB)` clearly shows the connection to slice.go now. I don't quite see the big picture yet, but now I know where to start if I wanted to dive deeper. :)

PR #842, opened 2023-12-29
Update ergochat-readline to 81f0f2329ad3 from cca60bf24c9f

wader:
[Source diff cca60bf24c9f..81f0f2329ad3](https://github.com/ergochat/readline/compare/vcca60bf24c9f..v81f0f2329ad3)

PR #846, opened 2024-01-05
bump readline to 0.1.0-rc1

slingamn:
ircdog PR here: https://github.com/ergochat/ircdog/pull/47

Excited for this! Are you seeing anything unexpected in testing? Have you been testing on Windows? (I did some Windows testing but I should probably do some more before the release.)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	// bump: ergochat-readline /github\.com\/ergochat\/readline .*-(.{12})$/ gitrefs:https://github.com/ergochat/readline.git|/master/|@commit|/(.{12})/
 	// bump: ergochat-readline command go get -d github.com/ergochat/readline@$LATEST && go mod tidy
 	// bump: ergochat-readline link "Source diff $CURRENT..$LATEST" https://github.com/ergochat/readline/compare/$CURRENT..$LATEST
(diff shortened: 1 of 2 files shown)


wader:
Hey, great! no haven't tested myself (i should really get a windows setup!) but i asked a guy who uses fq on windows here https://github.com/wader/fq/pull/842#issuecomment-1878381165 and it seems to be ok. Is there some specific areas your more worried about breaking on windows?

btw any opinion if fq should follow the master branch or stay with tagged releases?

slingamn:
I'd say tags unless there is a specific hotfix you want to pull in...generally my workflow is to do a final testing pass on release candidates before shipping them, so master is less tested.

Also I'm hoping that we need to do relatively little development work on the library going forward :-)

wader:
> I'd say tags unless there is a specific hotfix you want to pull in...generally my workflow is to do a final testing pass on release candidates before shipping them, so master is less tested.

👍

> Also I'm hoping that we need to do relatively little development work on the library going forward :-)

Agree, it seems to work fine and i don't i have any urgent features. Only thing i can think of now is  https://github.com/chzyer/readline/issues/212 that would be interesting but no way urgent.

PR #858, opened 2024-01-25
woff2: Add decoder

wader:
TODO:
- Worth adding dependencies for this?
- Split out ttf decoder somehow?


format/woff/woff2.go
```diff
@@ -0,0 +1,288 @@
+package woff
+
+// OpenType https://learn.microsoft.com/en-us/typography/opentype/
+
+import (
+	"bytes"
+	"io"
+	"time"
+
+	"github.com/dsnet/compress/brotli"
```
angea, line 10: 404 ?
not https://github.com/google/brotli ?

PR #876, opened 2024-02-10
fit: Formatted date/time description for timestamp fields

wader:


wader:
@mlofjard Not sure about this one, it says 2003, but it was 2023 right? i wonder also if there are for formats that could be made for fancy?

Also i'm a bit confused about format vs type. Ex "date_time" seems to be a type that gets set as a format in `LocalFieldDef`?

mlofjard:
Almost correct. =) The saved some space by using their own epoch from 31 Dec 1989 i think..  You need to add 631065600 seconds to s.Actual.

mlofjard:
Regarding Format vs Type, in the global definitions (those in messages_generated.go) Type can be "uint8" or "date_time", but when when the definition message arrives and we store the LocalFieldDef, the message would say something like { name: "timestamp", type: "uint32" }, so I store that type in LocalFieldDef.Type and "date_time" (from the global definition) in Format. Then when the data message arrives we know how to read the field from "Type" and how to display it from "Format".

mlofjard:
there is also a "local_date_time" that should have the same conversion

wader:
> Almost correct. =) The saved some space by using their own epoch from 31 Dec 1989 i think.. You need to add 631065600 seconds to s.Actual.

"They founded Garmin in 1989" huh :) found a reference for it in https://developer.garmin.com/fit/cookbook/decoding-activity-files/

And "local_date_time" seems to be in local timezone, i see there is some way of storing time_zone also, maybe for later. Think it's ok to assume local is also UTC for now?

mlofjard:
Yes, since it gets displayed with a Z anyway. local_date_time is just stored with more or less seconds.
`./fq '.data_records[].data_message | select(.local_timestamp)' format/fit/testdata/activity.fit` shows a message with both date_time and local_date_time.
local_timestamp should be `2023-09-03T09:06:00Z` since that is when I finished the race.

PR #878, opened 2024-02-12
fit: Skip array fields on pre read messages

mlofjard:
I also used your type->size idea for the expected sizes which simplified the switch statements alot.

### DIFF
--- a/format/fit/fit.go
+++ b/format/fit/fit.go
@@ -84,6 +84,26 @@ type localFieldDefMap map[uint64]map[uint64]mappers.LocalFieldDef
 type localMsgIsDevDef map[uint64]bool
 type valueMap map[string]valueType
 
+// expected size in bytes
+var expectedSizeMap = map[string]uint64{
+	"byte":    1,
+	"enum":    1,
+	"float32": 4,
+	"float64": 8,
+	"sint8":   1,
+	"sint16":  2,
+	"sint32":  4,
+	"sint64":  8,
+	"uint8":   1,
+	"uint8z":  1,
+	"uint16":  2,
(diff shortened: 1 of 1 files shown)


format/fit/fit.go
```diff
@@ -216,13 +236,24 @@ func fitDecodeDefinitionMessage(d *decode.D, drc *dataRecordContext, lmfd localF
 	}
 }
 
-func fieldUint(fieldFn func(string, ...scalar.UintMapper) uint64, expectedSize uint64, fDef mappers.LocalFieldDef, uintFormatter scalar.UintFn, fdc *fileDescriptionContext, valMap valueMap) {
+func readUint(d *decode.D, fDef mappers.LocalFieldDef, valMap valueMap) {
+	expectedSize := expectedSizeMap[fDef.Type]
+	if fDef.Size != expectedSize {
+		d.SeekRel(int64(fDef.Size) * 8) // skip over array types since they cannot be referenced by subfields
```
wader, line 242: How come this seek and the a seek in `fitDecodeDataMessage` does not cause "gaps", some other field cover that bit range?

wader:
+54 -44 😢 thought it would save lines 😬

format/fit/fit.go
mlofjard: This seek and the one on line 360 are just to keep the pos in sync since we only care about uint fields on the first pass. On line 363 we seek back to before the pre-read and read all the data with d.Field-methods.

mlofjard:
> +54 -44 😢 thought it would save lines 😬

You win some, you lose some =) At least it's less repeating code.

PR #885, opened 2024-02-27
pkg/cli/test_exp.sh: use command -v

thesamesam:
`which` is non-portable and not part of POSIX, but we can use `command -v` for the same effect.

Debian [0] and Gentoo [1] are both trying to drop which from their base system.

[0] https://lwn.net/Articles/874049/
[1] https://bugs.gentoo.org/646588


wader (approved):
Thanks! and thanks for the links

PR #893, opened 2024-03-10
nes: Add support for iNES/NES 2.0 ROM files

mlofjard:
This one has some "fun" stuff, like a toasm function for generating asm code, and tokitty for displaying graphics tiles in compatible terminals.


wader:
Thanks! Had a quick skimthru, overall good i think. Will look closer tonight or later in the week. One thing is the `toasm`/`tokitty` functions, maybe some kind of name spacing would be good to to know they are nes format related? or possible move them to documentation?

format/nes/nes.go
```diff
@@ -0,0 +1,339 @@
...
+		prgROMSize := romSize(prgLSB, 0, 16)
+		d.FieldValueUint("prg_rom_size_full", prgROMSize)
+		chrROMSize := romSize(chrLSB, 0, 8)
+		d.FieldValueUint("chr_rom_size_full", chrROMSize)
+
+		// byte 8
+		d.FieldU8("prg_ram_size", ramSizer(8))
+
+		// byte 9 - 15
+		d.FieldStr("crap", 7, encoding.Nop)
```
wader, line 222: Haven't seen `encoding.Nop` before. I wonder if it should be used in some other places in fq. I've thought about having maybe `d.FieldASCII` etc, i guss that would be nop?

format/nes/nes.go
```diff
@@ -0,0 +1,339 @@
...
+
+	d.Endian = decode.LittleEndian
+
+	d.FramedFn(16*8, func(d *decode.D) {
+		d.FieldStruct("header", func(d *decode.D) { decodeFileHeader(d, &nc) })
+	})
+
+	if nc.trainerSize > 0 {
+		d.FramedFn(int64(nc.trainerSize)*8, func(d *decode.D) {
+			d.FieldArray("trainer", func(d *decode.D) { decodePRGROM(d) })
```
wader, line 323: I think this and some below can be `d.FieldArray("trainer", decodePRGROM)` or want to keep for consistency with code above?

format/nes/nes.go
```diff
@@ -0,0 +1,339 @@
...
+	d.FramedFn(int64(nc.prgROMSize)*8, func(d *decode.D) {
+		d.FieldArray("prg_rom", func(d *decode.D) { decodePRGROM(d) })
+	})
+
+	d.FramedFn(int64(nc.chrROMSize)*8, func(d *decode.D) {
+		d.FieldArray("chr_rom", func(d *decode.D) { decodeCHRROM(d) })
+	})
+
+	if d.BitsLeft() > 0 {
+		d.FieldRawLen("the_rest", d.BitsLeft())
```
wader, line 336: Is it know as this? :) or could be "unknown"?

wader:
Noticed some possible fishy thing in `format/nes/testdata/nestest.fqtest` some text gets decode as op codes?

Screenshot 2024-03-11 at 18 10 06

format/nes/nes.go
```diff
@@ -0,0 +1,339 @@
...
+	mapperLSB := d.FieldU4("mapper_lsb")
+
+	// byte 7 (half)
+	consoleType := d.FieldU2("console_type", consoleTypeMapper)
+	nes20 := d.FieldU2("nes20_identifier", bitDesc("iNES", "NES 2.0"))
+
+	if nes20 == 2 { // NES 2.0
+		// byte 7 second half and byte 8
+		mapperMSB := d.FieldU8("mapper_msb")
+		d.FieldValueUint("mapper_full", mapperLSB+(mapperMSB<<4))
```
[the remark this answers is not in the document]
mlofjard: It's not an adress. Just an integer identifier

format/nes/nes.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func getShiftSizeMapper() scalar.UintFn {
+	return scalar.UintFn(func(s scalar.Uint) (scalar.Uint, error) {
+		if s.Actual > 0 {
+			s.Sym = 64 << s.Actual
+		}
+		return s, nil
+	})
+}
```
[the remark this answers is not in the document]
mlofjard: Yes, that looks much nicer. I had a function there before and it took the "64" in as an argument. But then the linter told me I was being dumb since I called it with "64" every time. =)

format/nes/nes.go
mlofjard: I was looking for an ASCII, but found this first, and since it is just padding that people sometimes puts text in I figured it was good enough.

format/nes/nes.go
mlofjard: Agree. There were more arguments before, so this is just leftover code smell =)

format/nes/nes.go
mlofjard: Could just as well be unknown. Or maybe I should just throw an error at this point since there isn't supposed to be anything beyond the CHR ROM.

mlofjard:
> Noticed some possible fishy thing in `format/nes/testdata/nestest.fqtest` some text gets decode as op codes?

Yup, that can happen. Since the PRG ROM is in fixed 16KiB blocks, it's very unlikely that all the data in it is actual code. And since the op code table contains all 256 entries (with the unofficial op codes) I have no way of knowing if a byte is an actual op code, added strings or random memory "junk" =)  I wrote a note about it in the limitations section of the docs.

mlofjard:
> Thanks! Had a quick skimthru, overall good i think. Will look closer tonight or later in the week. One thing is the `toasm`/`tokitty` functions, maybe some kind of name spacing would be good to to know they are nes format related? or possible move them to documentation?

Since they live in the nes.jq file, are they not scoped implicitely?

format/nes/nes.go
wader: Aha i see, makes sense

format/nes/nes.go
```diff
@@ -0,0 +1,329 @@
...
+func decodePRGROM(d *decode.D) {
+	for !d.End() {
+		peek := d.PeekUintBits(8)
+		peekTyp := opMap[peek].Type
+		peekArgL := ArgLength(peekTyp)
+
+		if d.BitsLeft() < int64(1+peekArgL)*8 {
+			d.FieldRawLen("junk", d.BitsLeft())
+		} else {
+
```
wader, line 237: Spurious new line

format/nes/nes.go
```diff
@@ -0,0 +1,329 @@
...
+func decodeCHRROM(d *decode.D) {
+	for !d.End() {
+		d.FieldStruct("tile", func(d *decode.D) {
+			lsbs := d.PeekBytes(8)
+			d.SeekRel(64)
+			msbs := d.PeekBytes(8)
+			d.SeekRel(-64)
+			d.FieldU64("data1")
+			d.FieldU64("data2")
```
wader, line 267: Could this be two FieldU64 calls?
```
lsbs := d.FieldU64("data1")
msbs := d.FieldU64("data2")
```

format/nes/nes.go
wader: I see, in most format decoders if there is some clear "end" signal i've usually just finish successfully and if there is something trailing there will be gap fields added automatically for it.

wader:
> > Thanks! Had a quick skimthru, overall good i think. Will look closer tonight or later in the week. One thing is the `toasm`/`tokitty` functions, maybe some kind of name spacing would be good to to know they are nes format related? or possible move them to documentation?
> 
> Since they live in the nes.jq file, are they not scoped implicitely?

All format *.jq files will be included globally no matter format being decoded atm. There is hacky way to have functions that will be overloaded based on input format https://github.com/wader/fq/blob/master/format/bson/bson.go#L24 but not sure it make sense in this case?

format/nes/nes.go
mlofjard: Maybe, if I add another loop to shift it down to 8 bytes again. "data1" is a 8x8 1bit image. I'll give it a try.

format/nes/nes.go
mlofjard: wader - mlofjard - linter // 1 - 0 - 0
Nice catch =)

mlofjard:
> > > Thanks! Had a quick skimthru, overall good i think. Will look closer tonight or later in the week. One thing is the `toasm`/`tokitty` functions, maybe some kind of name spacing would be good to to know they are nes format related? or possible move them to documentation?
> > 
> > 
> > Since they live in the nes.jq file, are they not scoped implicitely?
> 
> All format *.jq files will be included globally no matter format being decoded atm. There is hacky way to have functions that will be overloaded based on input format https://github.com/wader/fq/blob/master/format/bson/bson.go#L24 but not sure it make sense in this case?

Ah ok, then I get it. Do you think `to_nes_asm` and `to_nes_kitty` would suffice our should I be more specific? `from_nes_prg_to_asm`?

format/nes/nes.go
mlofjard: Made a bit of a bigger rewrite of the pixel combining code. It got rid of all the bitshifting and masking. Had to do some string sliceing and reversing instead due to endianess, but no more "peek and seek" =)

format/nes/nes.go
```diff
@@ -0,0 +1,325 @@
...
+
+//go:embed nes.jq
+//go:embed nes.md
+var nesFS embed.FS
+
+func init() {
+	interp.RegisterFormat(
+		format.NES,
+		&decode.Format{
+			Description: "NES 2.0",
```
wader, line 24: Wonder if it can be a bit more verbose like "NES 2.0 cartridge ROM format"?

format/nes/nes.jq
```diff
@@ -0,0 +1,9 @@
+def toasm:
+  select(.op_code and (.args // "")) | "\(.op_code) \(.args // "")";
+def tokitty(size):
+  def _sub: gsub("00"; "\u0022\u0022\u0022") | gsub("01"; "\u007F\u0022\u0022") | gsub("10"; "\u0022\u007F\u0022") | gsub("11"; "\u0022\u0022\u007F") | gsub(" "; "");
+  (select(.combined)
+  | "\(.combined | _sub)")
+  | "\\e_Ga=T,f=24,s=8,v=8,c=\(2*size),r=\(1*size);" + @base64 + "\\e\\\\\\n";
```
wader, line 7: I usually use "haskell"-ish jq style once jq program gets large, maybe something like:
```jq
def tokitty($size):
  def _sub:
    ( gsub("00"; "\u0022\u0022\u0022")
    | gsub("01"; "\u007F\u0022\u0022")
    | gsub("10"; "\u0022\u007F\u0022")
    | gsub("11"; "\u0022\u0022\u007F")
    | gsub(" "; "")
    );
  ( ( select(.combined)
    | "\(.combined | _sub)"
    )
  | "\\e_Ga=T,f=24,s=8,v=8,c=\(2*size),r=\(1*$size);" + @base64 + "\\e\\\\\\n"
  );
```

Is `.combined` always set for a `.chr_rom` struct? maybe can skip `select`?

A bit esoteric but size argument should probably be `$size` so that it is a value binding instead of a lambda. i.e. for `iskitty(1,2,3)` it should behave as `(1,2,3) as $a | iskitty($a)` not as `"...r=\1(1*(1,2,3));"` i guess?

Still debating what to about format specific addition functions. There is a bunch for mp4 and matroska that i've thought maybe move out some `mp4.jq`/`matroska.jq` library files and have in the wiki etc. And usage would then be something like `fq -L . 'include "mp4"; mp4_blabla' file.mp4`. Another option could be to look into jq modules and have namespace `mp4::blabla`, but i haven't used modules much yet.

Another concern is that currently to including of additinal jq code is not very performant, the AST will get cached but they will be compiled to byte code on each eval.

format/nes/nes.go
```diff
@@ -0,0 +1,325 @@
...
+		runes[i], runes[j] = runes[j], runes[i]
+	}
+	return string(runes)
+}
+
+func decodeTilePart(d *decode.D) scalar.Uint {
+	a, err := d.TryU64()
+	if err != nil {
+		panic(decode.IOError{Err: err, Name: "tile", Op: "Uint", Pos: d.Pos()})
+	}
```
wader, line 270: You want to throw tile specific panic error? if you do `a := d.U64()` it will also panic behind the scene. If you want custom fatal error you can use `d.Fatalf` or for non-fatal (can be skipped with `-o force=true`) use `d.Errorf`,

mlofjard: Yeah, this looks like late night brain fog to me. =) I must have been deep into decode_gen.go when I wrote this.

format/nes/nes.jq
mlofjard: I've prefixed them with `nes_` for now.

wader: 👍 let embed them for now, maybe in the future we move them out or make it more efficient somehow

format/nes/nes.go
wader: No worries :) i've lost count of times i've done something but then realised that i forgot i had already prepared or fix something month ago

format/nes/nes.go
```diff
@@ -0,0 +1,322 @@
...
+	// byte 6
+	d.FieldU1("nametable_layout", scalar.UintMapSymStr{0: "vertical", 1: "horizontal"})
+	d.FieldBool("battery", scalar.BoolMapDescription{false: "Not present", true: "Present"})
+	trainer := d.FieldBool("trainer", scalar.BoolMapDescription{false: "Not present", true: "Present"})
+	d.FieldBool("alternative_nametables")
+	mapperLSB := d.FieldU4("mapper_lsb")
+
+	// byte 7 (half)
+	consoleType := d.FieldU2("console_type", consoleTypeMapper)
+	nes20 := d.FieldU2("nes20_identifier", scalar.UintMapDescription{0: "iNES", 1: "NES 2.0"})
```
wader, line 148: Maybe battery, trainer, console_type and nes20_identifier should have symbolic names and maybe sometimes a description if symbolic value is not descriptive enough?

wader:
Maybe rename the test files after function name change?

Had some tiny comment but otherwise looks good it think

format/nes/nes.jq
```diff
@@ -0,0 +1,15 @@
...
+def nes_tokitty($size):
+  def _sub:
+    ( gsub("00"; "\u0022\u0022\u0022")
+    | gsub("01"; "\u007F\u0022\u0022")
+    | gsub("10"; "\u0022\u007F\u0022")
+    | gsub("11"; "\u0022\u0022\u007F")
+    | gsub(" "; "")
+    );
+  ( "\(.combined | _sub)"
+  | "\\e_Ga=T,f=24,s=8,v=8,c=\(2*$size),r=\(1*$size);" + @base64 + "\\e\\\\\\n"
```
wader, line 12: I swish jq had raw strings, fq actually does using <code>\`...\`</code> but not sure it's a good idea to use them in *.jq files

If you want to use @format + interpolation you can do probably do something like:
```jq
@base64 "\\e_Ga=T,f=24,s=8,v=8,c=\(2*$size),r=\(1*$size);\(.combined | _sub)\\e\\\\\\n"
```
... but not sure if more readable :)

mlofjard:
I found some more fun endianness bugs that I need to sort out (and create tests for), so there might not be any more commits before the weekend..

format/nes/nes.jq
wader: 👍

wader:
> I found some more fun endianness bugs that I need to sort out (and create tests for), so there might not be any more commits before the weekend..

Aha, no stress! will spend some time on trains this weekend so maybe will get some quality time with fq :)

wader (approved):
Looks good! let me know if your ready

mlofjard:
> Looks good! let me know if your ready

I've rebased on current master and squashed all my fixups, so I'm fine with merging as soon as CI is done.

wader:
Thanks!

Any other decoders in the works? i should really try finish up some decoders i have laying around

mlofjard:
> Thanks!
> 
> Any other decoders in the works? i should really try finish up some decoders i have laying around

Thank you as well. It's a fun project to contribute to.

I'm not sure what's next. I might look into some other ROM formats, like SNES or GB, in the future. But I'm also interested in what goes inside an EDID binary, since I had to patch mine to get 120Hz refresh rate working on my laptop. =)

wader:
> > Thanks!
> > Any other decoders in the works? i should really try finish up some decoders i have laying around
> 
> Thank you as well. It's a fun project to contribute to.

🥳 
 
> I'm not sure what's next. I might look into some other ROM formats, like SNES or GB, in the future. But I'm also interested in what goes inside an EDID binary, since I had to patch mine to get 120Hz refresh rate working on my laptop. =)

Nice! GB bring back memories, i wrote a gameboy emulator many years ago https://github.com/wader/maggie :)

EDID looks like it would be quite straight forward, little logic but lots of mapping!

PR #923, opened 2024-04-10
chore: fix function name in comment

loselarry:


wader:
Yeap, agree, maybe a linter that for each public function asserts that a doc exist and starts with the function name. Anyway, thanks for the fix!

PR #975, opened 2024-08-01
tzx: Add suport for ZX Spectrum TZX and TAP files

mrcook:
For those not familiar with the ZX Spectrum, this was a popular 1980's home micro in Europe and various other countries, and sold in excess of 5 million units. Most software was distributed on cassette tapes and these were encoded as TZX or TAP formats for use in emulators.

This PR adds decoders for these formats based on the _specification_, and also my previous efforts in writing a TZX/TAP decoder.

The test file includes a simple BASIC program and was exported from the FUSE emulator -- the TZX was manually edited to add some "archive info" (metadata).

Perhaps in a future update I could add functionality to decode the BASIC programs.


format/tap/testdata/basic_prog1.fqtest
```diff
@@ -0,0 +1,21 @@
+$ fq -d tap dv basic_prog1.tap
```
wader, line 1: Possible to add some instruction how the tap and tzx test files were created? if very simple just a comment in the .fqtest-file otherwise maybe a README.md in testdata?

format/tap/tap.md
```diff
@@ -0,0 +1,12 @@
...
+A TAP file is simply one data block or a group of 2 or more data blocks, one
+followed after the other. The TAP file may be empty.
+
+You will often find this format embedded inside the TZX tape format.
+
+The default file extension is `.tap`.
+
+### References
+
+- https://worldofspectrum.net/zx-modules/fileformats/tapformat.html
```
wader, line 12: Maybe you want to add a author section to tap.md and tzx.md? ex https://github.com/wader/fq/blob/master/format/fit/fit.md?plain=1#L14-L15

format/tzx/tzx.go
```diff
@@ -0,0 +1,655 @@
...
+			// NOTE: remove these next 4 fields from the length so
+			// the data size is calculated correctly
+			length -= 2 + 3 + 1 + 4
+
+			// Pause after this block (in ms)
+			d.FieldU16("pause")
+			// Sampling rate
+			d.FieldU24("sample_rate")
+			// Compression type
+			d.FieldU8("compression_type", scalar.UintMapSymStr{0x01: "RLE", 0x02: "Z-RLE"})
```
wader, line 168: I've usually tried to keep symbol mapped strings to be jq-ish, that is snake_case, but if the format spec etc is very clear about the naming then keep it think.

wader (commented):
Hey, thanks! it's a bit late in sweden so i've only done quick skim through but it looks very good 👍 will have a closer look tomorrow.

Is there some large archive of tap and txz files one could stress test the decoders on?

format/tap/testdata/basic_prog1.fqtest
```diff
@@ -0,0 +1,21 @@
...
+0x10|      26 00                                    |  &.            |        program_length: 38 0x12-0x14 (2)
+0x10|            01                                 |    .           |        checksum: 0x1 0x14-0x15 (1)
+    |                                               |                |    [1]{}: block 0x15-0x3f (42)
+0x10|               28 00                           |     (.         |      length: 40 0x15-0x17 (2)
+    |                                               |                |      data{}: 0x17-0x3f (40)
+0x10|                     ff                        |       .        |        flag: "standard_speed_data" (255) 0x17-0x18 (1)
+0x10|                        00 0a 14 00 20 f5 22 66|        .... ."f|        data: raw bits 0x18-0x3e (38)
+0x20|71 20 69 73 20 74 68 65 20 62 65 73 74 21 22 0d|q is the best!".|
+0x30|00 14 0a 00 ec 31 30 0e 00 00 0a 00 00 0d      |.....10.......  |
+0x30|                                          b6|  |              .||        checksum: 0xb6 0x3e-0x3f (1)
```
wader, line 21: Add a `$ fq -d tap -o read_one_block=true dv basic_prog1.tap` test here?

wader:
Failures `all.fqtest` i think is just to run tests with `-update`.

The failure in the zip tests i think is about that the tap decoder is in the probe group and succeeds on empty input, so maybe remove from probe group, if it's shaky to probe, or maybe make sure that at least one block was decoded? but it seems there is no real signature for a block so might not be possible? 🤔

mrcook:
> Is there some large archive of tap and txz files one could stress test the decoders on?

Hi Matthias, the go to place these days is https://spectrumcomputing.co.uk. I've tested over 50 tape images myself: mostly TZX but a dozen or so TAPs too.

Most tapes seems to be just the _Standard Speed Data_ and _Turbo Speed Data_ blocks, and I've struggled to find many TZX with the more interest blocks.

There are some collections available on the Internet Archive, so I plan to do some bulk testing at some point.

mrcook:
> The failure in the zip tests i think is about that the tap decoder is in the probe group and succeeds on empty input (`for !d.End() { ... }` does no iteration), so maybe remove from probe group, if it's shaky to probe, or maybe make sure that at least one block was decoded? but it seems there is no real signature for a block so might not be possible? 🤔

Yep, and in fact the specs state: "The TAP file may be empty. Then it has a size of 0 bytes".

I hadn't looked into what that probe group does, I just started with a copy of the format/nes decoder 🙃 I'll try removing it and see if that works.

format/tap/testdata/basic_prog1.fqtest
mrcook: Hmm, on running this all blocks are still being decoded. I'll take another look at this tonight.

wader:
> There are some collections available on the Internet Archive, so I plan to do some bulk testing at some point.

Great! if you end up writing some script or has some snippet to do it you can include that in testdata/README.md also. I've usually tried to document test data well so that me, you or someone in the future as some resonable chance at updated them :)

wader:
> Yep, and in fact the specs state: "The TAP file may be empty. Then it has a size of 0 bytes".
> 
> I hadn't looked into what that probe group does, I just started with a copy of the format/nes decoder 🙃 I'll try removing it and see if that works.

Aha i see. Being part of the "probe group" just means that fq will, if no format or group is specific with `-d`, try to decode using all format in that group and the first format to succeed will be selected. So that is why it's important to not succeed on empty input and also try to fail fast if part of the probe group. Also some archive formats like zip uses the probe group to do nested decoding.

About nes decoder: Oh then i should probably have a look at it so it does not have the same issue.

format/tap/testdata/basic_prog1.fqtest
mrcook: After a quick debug I notice that the `ReadOneBlock` flag is not being set here:

```
var ti format.TAP_In
d.ArgAs(&ti)
```

Is my implementation missing something?

format/tap/tap.go
```diff
@@ -0,0 +1,146 @@
...
+
+//go:embed tap.md
+var tapFS embed.FS
+
+func init() {
+	interp.RegisterFormat(
+		format.TAP,
+		&decode.Format{
+			Description: "TAP tape format for ZX Spectrum computers",
+			DecodeFn:    tapDecoder,
```
wader, line 24: Add this and i think the `read_one_block` option should work:
```go
			DefaultInArg: format.TAP_In{
				ReadOneBlock: false,
			},
```
It's how the fq internals know the type and also default values

format/tap/testdata/basic_prog1.fqtest
wader: Yeap see my comment for `tap.go`

format/tap/tap.go
mrcook: I did initially think to add this, but then figured, it's false by default so no need to add it :)

wader: Aha :) i've thought about making `Format` generic somehow then the in/out types could be passed as use zero value... but maybe not worth it? maybe will refactor this in the future

format/tzx/testdata/README.md
```diff
@@ -0,0 +1,28 @@
+### basic_prog1.tzx
+
+The `basic_prog1.tzx` test file was created directory from the FUSE emulator.
+
+Inside the emulated ZX Spectrum a BASIC program was created:
+
+```
+10 PRINT "fq is the best!"
```
wader, line 8: 🥳

format/tzx/tzx.go
```diff
@@ -0,0 +1,655 @@
...
+		// consists of 8063 pulses if the first data byte (the flag byte)
+		// is < 128, 3223 otherwise.
+		0x10: func(d *decode.D) {
+			// Pause after this block (ms.) {1000}
+			d.FieldU16("pause")
+
+			// A single TAP Data Block
+			peekBytes := d.PeekBytes(2)                              // get the TAP data block length
+			length := uint16(peekBytes[1])<<8 | uint16(peekBytes[0]) // bytes are stored in LittleEndian
+			length += 2                                              // include the two bytes for this value
```
wader, line 70: Is the `ReadOneBlock` option or could the length*8 limit and the end check be enough?

mrcook: Right! In retrospect it even makes sense for these embedded tap blocks would be an array. As the TAP spec states: "a TAP file is simply one datablock or a group of 2 or more datablocks".

format/tzx/tzx.go
```diff
@@ -0,0 +1,655 @@
...
+		// use a utility to join the two files and ensure that they are both
+		// of the higher version number.
+		0x5A: func(d *decode.D) {
+			// Value: { "XTape!",0x1A,MajR,MinR }
+			// Just skip these 9 bytes and you will end up on the next ID.
+			d.FieldRawLen("value", int64(9*8))
+		},
+	}
+
+	blockType := d.FieldU8("block_type", blockTypeMapper)
```
wader, line 411: Maybe can be just "type" if it's inside a block struct?

format/tap/tap.go
```diff
@@ -0,0 +1,139 @@
...
+func decodeTapBlock(d *decode.D) {
+	// Length of the following data.
+	length := d.FieldU16("length")
+
+	// read header, fragment, or data block
+	switch length {
+	case 0:
+		// fragment with no data
+	case 1:
+		d.FieldRawLen("data", 8)
```
wader, line 56: Possible future change: Are these data fragments (and data below) something that can be joined into something decodable? if so fq can append things together and decode a "sub"-buffer of some format or just provide a large raw field

format/tap/tap.go
```diff
@@ -0,0 +1,139 @@
...
+		// In case of a SCREEN$ header = 16384.
+		d.FieldU16("start_address", scalar.UintHex)
+		//	UnusedWord: 32768.
+		d.FieldU16("unused")
+	default:
+		d.Fatalf("invalid TAP header type, got: %d", dataType)
+	}
+
+	// Simply all bytes XORed (including flag byte).
+	d.FieldU8("checksum", scalar.UintHex)
```
wader, line 122: fq has some support for checksums if you like to add it https://github.com/wader/fq/blob/cdba38d067b67d22936299baaf8cc415042f4a43/format/png/png.go#L232-L233

wader:
Just some minor comments that you might want to do, otherwise i think it look good as it is now. And can always do more changes later on, i like to iterate and not let PRs drag on too long. Let me know when you feel your happy with the code and testing.

BTW what's your use case for this? just curious how the formats work or using to do some aggregations or querying for ROMs features etc?

mrcook:
> Just some minor comments that you might want to do, otherwise i think it look good as it is now. And can always do more changes later on, i like to iterate and not let PRs drag on too long. Let me know when you feel your happy with the code and testing.

I agree that iterating is better than having long running PRs. Your new comments can certainly be considered improvements rather than requirements.

Whatever you think is best. :)


> BTW what's your use case for this? just curious how the formats work or using to do some aggregations or querying for ROMs features etc?

Exactly that; extracting metadata, features, etc. The archive info (title, publisher, etc.) is the obvious one, but also the more interesting blocks like Hardware Type. Possibly this could be used for verifying/updating the https://github.com/zxdb/ZXDB

I already have the core of a BASIC decoder written, so in the future I'd also like to add that.

wader:
> > Just some minor comments that you might want to do, otherwise i think it look good as it is now. And can always do more changes later on, i like to iterate and not let PRs drag on too long. Let me know when you feel your happy with the code and testing.
> 
> I agree that iterating is better than having long running PRs. Your new comments can certainly be considered improvements rather than requirements.
> 
> Whatever you think is best. :)

The smaller things maybe in this PR, the larger about fragments etc maybe later if you feel motivated.

> > BTW what's your use case for this? just curious how the formats work or using to do some aggregations or querying for ROMs features etc?
> 
> Exactly that; extracting metadata, features, etc. The archive info (title, publisher, etc.) is the obvious one, but also the more interesting blocks like Hardware Type. Possibly this could be used for verifying/updating the https://github.com/zxdb/ZXDB

Sounds similar to my use case for media files, diff, aggregate and query for very things. But i'll say a majority of the decoders in fq exist mostly out of curiosity :)

> I already have the core of a BASIC decoder written, so in the future I'd also like to add that.

👍 I've done some experiments with decoding ISA:s to be something like a "structured  disassembler", see https://github.com/wader/fq/pull/215, maybe can give some inspiration how it might work.

mrcook:
> The smaller things maybe in this PR, the larger about fragments etc maybe later if you feel motivated.

I should be able to take a look at that tomorrow evening.

format/tap/tap.go
mrcook: It's not immediately obvious to me how I'm going implement this, so I won't get this done tonight.

wader: Ok but no worries, is up to you. Could be done later if it's something you want

wader:
test `-update` should fix CI error

format/tap/tap.go
mrcook: I thought I'd have a quick look before I go to bed and put something together.

wader:
Looks good. Ready to merge or you wanted to do some more testing?

mrcook:
I'd be happy with what we have here now, so yes, feel free to merge it.

mrcook:
Thanks Matthias! I also wanted to thank you for all the help in making my contribution better.

Last but no least, thanks for creating this great tool!!

wader:
Glad to hear that 😊 and looking forward to future contributions! and if you have questions about how to use fq in general i'm happy to help, always interesting and useful to know how ppl are using it.

PR #992, opened 2024-08-19
interp: Add bits_format=byte_array support

wader:
Represent binary values into an array of bytes when turn values into JSON


wader:
@mrcook

wader:
Still not sure if it should be `byte_array` or `bytes_array` 🤔

wader:
"byte array" seems to be more common when i search around a bit, going with that

wader:
@mrcook feel free to add some tzx docs using byte_array now if you want

PR #999, opened 2024-08-21
tzx: add readme instructions for bits_format=byte_array

mrcook:
A small update to the TZX/TAP formats readme, with reference to the new `bits_format=byte_array`.


wader:
Thanks! 👍

PR #1023, opened 2024-11-07
midi: minor fixups

twystd:
Hi,

This pull request implements some small cleanups for the midi plugin:
- adds (fairly minimal) package godoc
- reworks the SMPTE field decoding in the _MThd_ header to more closely align with the wording of the MIDI file spec.
- reworks the decoding of MIDI format 0 files to only decode the first _MTrk_ chunk with any additional (superfluous) _MTrk_ chunks being decoded as raw _data_. Again this more closely aligns with the MIDI file spec.


format/midi/constants.go
```diff
@@ -368,9 +370,9 @@ var framerates = scalar.UintMapSymUint{
...
-var fps = scalar.UintMapSymUint{
-	0xe8: 24,
-	0xe7: 25,
-	0xe6: 29,
-	0xe5: 30,
+var fps = scalar.SintMapSymUint{
+	-24: 24,
+	-25: 25,
+	-29: 29,
+	-30: 30,
```
wader, line 377: Huh what is this weird mapping? :) is stored as 2's compliment but the abs value is always used? 🤔

format/midi/midi.go
```diff
@@ -33,17 +48,31 @@ func init() {
...
+// decodeMIDI implements the MIDI file decoder.
+//
+// The decoder parses the file as a set of chunks, each comprising a 4 character tag
+// followed by a uint32 length field. The decoder parses the MTHd and MTrk MIDI chunks.
 func decodeMIDI(d *decode.D) any {
 	d.Endian = decode.BigEndian
 
 	// ... decode header
-	d.FieldStruct("header", decodeMThd)
+	format, _ := decodeMThd(d)
```
wader, line 59: Oh you also ran into the issue of wanting to use `d.Field...` but also return a value? i've thought about if generics could help but think i ran into some limitations

format/midi/constants.go
```diff
@@ -368,9 +370,9 @@ var framerates = scalar.UintMapSymUint{
 	3: 30,
 }
 
-var fps = scalar.UintMapSymUint{
-	0xe8: 24,
-	0xe7: 25,
-	0xe6: 29,
-	0xe5: 30,
+var fps = scalar.SintMapSymUint{
```
wader, line 373: I noticed above `manufacturers_extended`, maybe use camel case to be more like reste of go code? maybe there are others?

I've usually named these mappers `<thing>Names`, `<thing>Map` etc maybe good here also?

format/midi/testdata/Makefile
```diff
@@ -16,8 +16,10 @@ run: build
 	go run . -d midi dv format/midi/testdata/reference.mid
 
 debug: build
-	go run . -d midi dv format/midi/testdata/midi/smpte-timecode.mid
-	go run . -d midi dv format/midi/testdata/midi/format-0.mid
+	go run . -d midi dv format/midi/testdata/midi/format-0x.mid
+	go run . -d midi dv format/midi/testdata/midi/format-0y.mid
+	go test ./format -run TestFormats/midi/testdata/format-0x.fqtest
+	go test ./format -run TestFormats/midi/testdata/format-0y.fqtest
```
wader, line 22: Don't see these files anywhere or maybe it's go test framework that mangles the test names somehow?

format/midi/constants.go
twystd: Yup, welcome to the wonderful world of MIDI :-). 

However, fair point and I've remapped them to strings for clarity because 29 is not 29 - it's actually "_29.97 FPS and drop the occasional frame to keep things in sync_". **And** also discovered a bug in the _SMPTE offset_ event decoding while I was at it, so good call!

FYA: here's the relevant bit of the spec:

> If bit 15 of <division> is a one, delta-times in a file correspond to subdivisions of a second, in
> a way consistent with SMPTE and MIDI time code. Bits 14 thru 8 contain one of the four
> values -24, -25, -29, or -30, corresponding to the four standard SMPTE and MIDI time code
> formats (-29 corresponds to 30 drop frame), and represents the number of frames per second.
> These negative numbers are stored in two's complement form. The second byte (stored
> positive) is the resolution within a frame: typical values may be 4 (MIDI time code resolution),
> 8, 10, 80 (bit resolution), or 100. This system allows exact specification of time-code-based
> tracks, but also allows millisecond-based tracks by specifying 25 frames/sec and a resolution of
> 40 units per frame. If the events in a file are stored with bit resolution of thirty-frame time
> code, the division word would be E250 hex.

and:

> The new counting system counts from 0 to 29 for most seconds, but skips frame 0 and frame 1 for some 
> seconds, jumping directly from frame 29 of the previous second to frame 2 of the next second.  This counting 
> system is called “drop frame.”

format/midi/midi.go
twystd: &lt;grin&gt; I sat there for a whole minute wishing `d.FieldStruct` returned a value and then resigned myself to doing it the messy way.

format/midi/constants.go
twystd: Makes sense .. renamed throughout.

format/midi/testdata/Makefile
twystd: Gah! It's in the _debug_ target of the _Makefile_ - they were the initial files I was testing with. Have fixed it to use the renamed versions (_format-0-extra-tracks_ and _format-0-extra-tracks2_).

format/midi/midi.go
wader: Yeah sadly a bit messy and methods are not allowed to have type params in go :(

format/midi/constants.go
wader: Oh 29.97 i guess from the us electric grid? same thing shows up in other digital media formats :) i guess you could use sym float if you think ppl would like to use the value in queries etc.

wader:
Looks good to me! merge?

format/midi/constants.go
twystd: It's from film to NTSC (TV) conversion - there's a whole long writeup [here](https://finale3d.com/documentation/time-code-frame-rate) if you're interested.

re. sym float. I thought about it but the rate is actually 29.97002997 **and** drop a frame so however you code it you're going to end up writing a switch statement so clarity seemed the better way to go.

twystd:
re. merge. I don't have anything else to add so I'm good to merge if you're happy :-)

wader:
Thanks!

wader:
They say meowllo! image

twystd:
Awwwww .. they're beautiful :-) :-)

PR #1047, opened 2024-12-12
Replace `exp/slices` package with standard library

Juneezee:
The experimental functions in `golang.org/x/exp/slices` are now included in the standard library since Go 1.21 [^1].

[^1]: https://go.dev/doc/go1.21#slices

However, we cannot remove the `golang.org/x/exp` dependency yet because we still rely on the `golang.org/x/exp/constraints` package, which has not yet been included into the standard library (https://github.com/golang/go/issues/61914).

An alternative approach is to create our own implementation, e.g. `constraints.go` like this PR https://github.com/hashicorp/terraform-plugin-testing/pull/260 so we can remove the `golang.org/x/exp` dependency completely.

https://github.com/wader/fq/blob/4e3ab79695fbec3e7fa275cb5bacac8587d852c1/format/apple/loop_detector.go#L4

https://github.com/wader/fq/blob/4e3ab79695fbec3e7fa275cb5bacac8587d852c1/internal/mathx/num.go#L11

### DIFF
--- a/format/mp4/mp4.go
+++ b/format/mp4/mp4.go
@@ -19,12 +19,12 @@ import (
 	"cmp"
 	"embed"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/wader/fq/format"
 	"github.com/wader/fq/pkg/decode"
 	"github.com/wader/fq/pkg/interp"
-	"golang.org/x/exp/slices"
 )
 
 //go:embed mp4.jq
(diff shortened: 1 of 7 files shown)


wader:
Thanks 👍 yeah maybe would be nice to remove golang.org/x/exp dependency completely then. Feel free to amend that to this PR or create a new PR for it.

Juneezee:
> Thanks 👍 yeah maybe would be nice to remove golang.org/x/exp dependency completely then. Feel free to amend that to this PR or create a new PR for it.

Done! Please take a look again. Thanks :blush:

Issue #1087, opened 2025-02-27
FQ fails to parse YAML files containing multiple documents


visualphoenix:
#### What version are you using (`fq -v`)?

<pre>
$ fq -v
0.13.0 (darwin arm64 go1.23.5)
</pre>

#### How was `fq` installed?

FQ was installed with Nix on macOS.

#### Can you reproduce the problem using the latest release or master branch?

I have verified that this issue occurs in version 0.13.0. I haven't checked the master branch, but I can try if needed.

#### What did you do?

I attempted to parse a YAML file containing multiple documents using `fq`.

To demonstrate:
```
$ cat b
---
hi: bye
---
doesnt: work

$ cat b | fq .
```

#### What result did you expect?

I expected `fq` to parse the YAML file and output each document separately or in an array format, similar to how it handles a single document.

#### What did you see instead?

```
error: <stdin>: probe: failed to decode: try fq -d FORMAT to force format, see fq -h formats for list
```

wader:
Hi, thanks for the report! That should be possible to fix but i'm a bit unsure how it should work. I did a draft PR #1088 that decodes all documents into an array but then these two yaml files will produce the same jq value:
```yaml
---
hi: bye
---
doesnt: work
```
```yaml
- hi: bye
- doesnt: work
```
Maybe not a big deal. But maybe the yaml format,`to_yaml` and `from_yaml` should have an options to force multi document mode?

visualphoenix:
In my case I’m dealing with kubernetes manifests and helm template output - eg: https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.1/experimental-install.yaml

It’s useful to do analysis on kinds. Right now I have to use yq to parse the multi document yaml into json and then use fq.

Eg: cat foo.yaml | yq e -o json | fq .

Would be nice to get rid of yq entirely. That’s my last missing use case

visualphoenix:
It would also be nice if it preserved ordering eg 

cat experimental-install.yaml | fq -rc to_yaml

Was a multi document output ideally with stable key ordering. I know that might be tough with go.

wader:
Now the draft PR has has a `multi_document` options for both decode and encode to force multi document mode.

If you have go installed you should be albe to try it out using:
```sh
$ go run github.com/wader/fq@yaml-multi-document -nr '[1,2,3] | to_yaml({multi_document: true})'
1
---
2
---
3

```

> Was a multi document output ideally with stable key ordering. I know that might be tough with go.

[...]

PR #1088, opened 2025-02-27
yaml: Support multi document decode

wader:
Related to #1087

TODO:
Support encode via option?


visualphoenix:
you might consider https://github.com/kubernetes/apimachinery/tree/master/pkg/util/yaml or https://github.com/kubernetes-sigs/yaml instead of gopkg.in/yaml.v3 which (links to https://github.com/go-yaml/yaml/tree/v3.0.1 ) hasnt been updated in 3 years

wader:
> you might consider https://github.com/kubernetes/apimachinery/tree/master/pkg/util/yaml 

Seem to be decode only?

> or https://github.com/kubernetes-sigs/yaml

This one might be interesting. Looks like a wrapper and fork of go-yaml?

> instead of gopkg.in/yaml.v3 which (links to https://github.com/go-yaml/yaml/tree/v3.0.1 ) hasnt been updated in 3 years

Is there some particular bug or issue with the current yaml dependency? but yes if there is a more active yaml implementation it would be nice to switch

PR #1108, opened 2025-03-24
Correct a few spelling mistakes.

01mf02:


wader:
Thanks! and sorry about the state of the usage docs, a bit messy and lots of todo:s, they might be a bit out of date also.

01mf02:
@wader, it's always better to have any kind of documentation than none. :) I already learnt quite a few things from reading your docs, so do not worry about them not being perfect. ;) I have to thank you for writing things up!

wader:
👍 there is also https://github.com/wader/fq/blob/master/doc/dev.md but that one is even worse state :)

01mf02:
Thanks for pointing it out to me!

PR #1110, opened 2025-03-27
mp4: uuid: decode tfxd and tfrf boxes

rafalborczuch:
Decoding support for TFXD and TFRF boxes based on the latest Smooth Streaming [specification](https://winprotocoldocs-bhdugrdyduf5h2e4.b02.azurefd.net/MS-SSTR/%5bMS-SSTR%5d.pdf)  (v20241119).


format/mp4/boxes.go
```diff
@@ -1367,8 +1367,45 @@ func decodeBox(ctx *decodeContext, d *decode.D, typ string) {
 		d.FieldU32("h_spacing")
 		d.FieldU32("v_spacing")
 	case "uuid":
-		d.FieldRawLen("uuid", 16*8, scalar.RawUUID, uuidNames)
-		d.FieldRawLen("data", d.BitsLeft())
+		uuidBR := d.FieldRawLen("uuid", 16*8, scalar.RawUUID, uuidNames)
```
wader, line 1370: I wonder if a user wants sym to always be the dashed uuid string and the names mapping to set description? atm sym is only used and will be a dashed uuid string as fallback, maybe ok also? 🤔

format/mp4/boxes.go
```diff
@@ -1367,8 +1367,45 @@ func decodeBox(ctx *decodeContext, d *decode.D, typ string) {
...
+							d.FieldU64("absolute_time")
+							d.FieldU64("duration")
+						} else {
+							d.FieldU32("absolute_time")
+							d.FieldU32("duration")
+						}
+					})
+				}
+			})
+		case uuid == nil:
```
wader, line 1404: Think this can't happen?

format/mp4/boxes.go
```diff
@@ -1367,8 +1367,45 @@ func decodeBox(ctx *decodeContext, d *decode.D, typ string) {
...
+		// TODO: make nicer
+		uuid := d.ReadAllBits(uuidBR)
+
+		switch {
+		case bytes.Equal(uuid, uuidTfxdBytes[:]):
+			version := d.FieldU8("version")
+			d.FieldU24("flags")
+			d.FieldStruct("fragment", func(d *decode.D) {
+				if version == 1 {
+					d.FieldU64("absolute_time")
```
wader, line 1380: Does these timestamp have an epoch? if so they could use something like `scalar.UintActualUnixTimeDescription(time.Second, time.RFC3339)` but don't see anything in the spec

format/mp4/testdata/tfxd_tfrf.mp4.fqtest
```diff
@@ -0,0 +1,31 @@
+$ fq dv -d mp4 -o force=true tfxd_tfrf.mp4
```
wader, line 1: 🥳 thanks for adding a test!

wader (commented):
Hi, thanks! looks good 👍 just a few thoughts

format/mp4/boxes.go
```diff
@@ -1367,8 +1367,45 @@ func decodeBox(ctx *decodeContext, d *decode.D, typ string) {
...
 		d.FieldU32("v_spacing")
 	case "uuid":
-		d.FieldRawLen("uuid", 16*8, scalar.RawUUID, uuidNames)
-		d.FieldRawLen("data", d.BitsLeft())
+		uuidBR := d.FieldRawLen("uuid", 16*8, scalar.RawUUID, uuidNames)
+		// TODO: make nicer
+		uuid := d.ReadAllBits(uuidBR)
+
+		switch {
+		case bytes.Equal(uuid, uuidTfxdBytes[:]):
```
wader, line 1375: If you want you could add comment with spec name/version/url etc where the structures come from

format/mp4/boxes.go
rafalborczuch: I kept it the same way as it used to be. I think it's fine.

format/mp4/boxes.go
rafalborczuch: It doesn't have any epoch. It represents time in timescale increments for the track.

format/mp4/boxes.go
rafalborczuch: I copied this block from line 1337. You are right, `uuid` can't be `nil` here.

wader:
Thanks!

PR #1130, opened 2025-04-19
fix typo

retokromer:
- fix typo
- use US English everywhere


wader:
Thanks! not sure why but don't think i've ever managed to spell behavior correctly at first try :)

PR #1155, opened 2025-07-01
Add Manual to TODO.md

nichtich:


wader:
Good idea. Interactive as in a web page etc? i've started to add a man page a couple of times https://github.com/wader/fq/pull/698 maybe should continue on that

PR #1233, opened 2025-12-07
update EBML link

retokromer:
Update EPBL link to the published RFC 8794.


wader:
Thanks! btw is the URL to file used to generate part of the matroska decoder correct? see https://github.com/wader/fq/blob/master/format/matroska/ebml_matroska/ebml_matroska.go?

Issue #541, opened 2022-12-25
Add gentoo install instructions


wader:
What is the proper way install fq on gentoo? `emerge dev-util/fq`? package is available in "stable" etc now?

wader:
@thesamesam ^

thesamesam:
Sorry, I'd missed this! d9018fad22b31deb6959849584a5458bcd95117d indeed looks good :)

wader:
@thesamesam 👍 now worries :)

PR #1314, opened 2026-05-10
Add Gentoo install instructions

kimjune01:
Adds Gentoo install instructions to README per #541.

Package is available in stable via `emerge dev-util/fq`.


wader:
Thanks!

Issue #1318, opened 2026-05-28
[Feature Request]: Support `$ARGS`


AZMCode:
### What version are you using (`fq -v`)?

<pre>
$ fq -v
0.17.0 (linux amd64 go1.26.2)
</pre>

### How was fq installed?
Installed from `nixos-unstable`'s `fq` package.

```
$ nix-info -m
 - system: `"x86_64-linux"`
 - host os: `Linux 6.18.31, NixOS, 26.05 (Yarara), 26.05.20260515.d233902`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.34.7`
 - channels(root): `"nixos, nixpkgs"`
 - nixpkgs: `/nix/store/77dbgds155bbz3vd3qywq1sii07i5ljs-source`
```

### Can you reproduce the problem using the latest release or master branch?

Yes. Master as of time of writing (commit SHA `b3f8bd99c1e97736ea61f898ee81eac8b31d7206`)

```
$  fq -n '$ARGS' --arg avar A
error: arg: variable not defined: $ARGS
```

### What did you do?

I tried to use '$ARGS' in a filter.

```
$  fq -n '$ARGS' --arg avar A
error: arg: variable not defined: $ARGS
```

### What result did you expect?

Obtain an object/array containing information on all arguments passed, to generate arrays and structures from

### What did you see instead?
```
error: arg: variable not defined: $ARGS
```

### Extra Info
This feature is supported by `jq`, as well as positional arguments, but that's a different issue altogether.

wader:
Hey, yeap that seems like something fq should support

wader:
Sorry for the delay, a bit busy with other things. A simple fix for this is probably quite simple but maybe it would be good to also fix some other compatibility issues related to the arg-arguments.

```
# args.sh
JQ=$1

# --arg prio
$JQ -cn --arg a 1 --arg a 2 '$ARGS'

# --arg vs --jsonarg prio
$JQ -cn --argjson a 1 --arg a 2 '$ARGS'

# ARGS prio
$JQ -cn --arg ARGS 1 '$ARGS'

# --args collect non-dash-args
$JQ -cn --args '$ARGS' -n a -n b

# --args with -- collects all args
$JQ -cn --args -- '$ARGS' -n b -n

# --jsonargs collects non-dash-args
$JQ -cn --jsonargs '$ARGS' -n '"a"' -n '"b"'

# --jsonargs with -- collects non-dash-args
[...]

PR #1321, opened 2026-06-03
interp: Add $ARGS, --args and --jsonargs suppport

wader:
Also make --arg/---argjson priority more like jq
Mixing --args and --jsonargs is still a bit weird

Related to #1318

### DIFF
--- a/pkg/interp/args.jq
+++ b/pkg/interp/args.jq
@@ -31,7 +31,11 @@ def _args_parse($args; $opts):
         $r
       else
         if $arg == "--" then
-          $r | .rest += $args[1:]
+          if $r.positional then
+            $r | .parsed[$r.positional] += $args[1:]
+          else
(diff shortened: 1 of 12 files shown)


pkg/interp/args.jq
```diff
@@ -103,7 +113,14 @@ def _args_parse($args; $opts):
     | map({(.key): .value.default})
     | add
     );
-  _parse($args; _flagmap; {parsed: _defaults, rest: []});
+  _parse(
+    $args;
+    _flagmap;
+    { parsed: _defaults
+    , rest: []
+    , positionals: null
```
wader: Not needed

wader: Actually yes, keep to be explicit

wader: But rename

Issue #1371, opened 2026-08-21
Supporting more EBML-based formats


martinkirch:
Hello and first thanks for fq ! After spending too many hours against `hexdump` I was truly glad to find out this program.

I'm working on MOSAIC, [a BLOB container tailored for a source code archive](https://docs.softwareheritage.org/devel/swh-mosaic/). The format has been designed using EBML, the meta-language designed for matroska. As matroska was already supported by fq, I could add support for mosaic files quite easily - see [the comparison with my branch](https://github.com/wader/fq/compare/master...martinkirch:fq:nicer-mosaics?expand=1). The ability to list indexes with `fq '.mosaic.index[].idx_description'` is really cool!

I'm opening this issue to know if you could be interested in merging some of this patches. If yes, I should rework commits and MR, starting with one about `format/matroska`. Because we'd need at least to move a few functions to `format/matroska/ebml`, so they'll be easier to re-use for ebml, and check `ebml_doctype=="matroska"`. but there's more :

   - my current patch breaks matroska tests because I changed the way "singleton" elements were constantly wrapped in a struct called `element` ([here](https://github.com/wader/fq/compare/master...martinkirch:fq:nicer-mosaics?expand=1#diff-bb9b73d2531e948acb7db80f0d03628142ba4350a75d81c35ba68280f4d56735R255)). This allows for shorter and more explicit fq queries. Not sure if I should update tests or revert this change.
   - but maybe it would be better to add `format/ebml` entirely ? The difficulty of EBML is that it cannot be rigorously parsed without knowing the underlying schema, that is usually hardcoded and documented as XML. Digging this idea leads to two methods : either support generating `ebml.Element` at runtime from a user-provided XML (this idea goes beyond my current Go knowledge, though), or try a generalized-heuristic-EMBL-decoder by considering element data starting with two valid VInt might be a master element, all others would be typed binary.

Curious to have opinions on this !

wader:
> Hello and first thanks for fq ! After spending too many hours against `hexdump` I was truly glad to find out this program.

Glad to hear that! :) i've also spent quite a lot of time with hexdump and other similar tools (still do!)

[...]

PR #1375, opened 2026-08-25
Separate EBML from Matroska, add another EBML-based format: MOSAIC

martinkirch:
as discussed in #1371

the only modification to the Matroska parser is a check of the `doctype` field; without this, Matroska would probe positively any EBML file.


format/matroska/ebml_matroska/ebml_matroska_gen.go
```diff
@@ -1,8 +1,8 @@
-// Code below generated from ebml_matroska.xml
+// Code below generated from format/matroska/ebml_matroska/ebml_matroska.xml
```
wader: I guess this changed due to manual run of gen? if you do `make gogenerate` (that just runs `go generate -x ./...`) it will probably change back

format/mosaic/mosaic.md
```diff
@@ -0,0 +1,25 @@
+
```
wader, line 2: Could you format more like mp4.md or matroska.md for now? it's a good idea to be able to "append" more to the automatic decode examples section. Maybe if one have a "### Decode examples" section it will get appended in the help? 🤔  

The file will later be used to generate documentation (in next release both man page and html) that relies on sections and section depth.

format/mosaic/mosaic.md
```diff
@@ -0,0 +1,25 @@
...
+    $ fq '.mosaic.index[].idx_description' file.mosaic
+
+    # List only small objects
+    $ fq '.mosaic.tile[].object[] | select(.size < 1000)' file.mosaic
+
+    # Check if the reported size is correct
+    $ fq 'add(.mosaic.tile[].object[].size) == .mosaic.container_meta_data.objects_total_size.value' basic.mosaic
+
+    # Command above only works if the file does not use compression - check it with:
+    $ fq '.mosaic.container_meta_data.compression_method | d' basic.mosaic
```
wader: Possibly leave out `... | d`? or be fancy `fq -Vr '.mosaic.container_meta_data.compression_method.value' basic.mosic`?

format/mosaic/mosaic.go
```diff
@@ -0,0 +1,170 @@
...
+		format.MOSAIC,
+		&decode.Format{
+			Description: "MOdular Storage of Archived and Indexed Contents",
+			Groups:      []*decode.Group{format.Probe},
+			DecodeFn:    mosaicDecode,
+		})
+	interp.RegisterFS(mosaicFS)
+}
+
+func decodeMaster(d *decode.D, bitsLimit int64, elm *ebml.Master) {
```
wader, line 28: This is much nicer and produces much nicer decode tree than the matroska code. I'm not sure why i didn't opt for a similar structure, maybe was a bit unsure about uniqueness restrictions in ebml (or was spooked by https://www.rfc-editor.org/rfc/rfc8794.html#name-identically-recurring-eleme, but not sure i've even seen it in practice?). Will probably convert matroska later to something similar, i did a quick hack using your code it seems to work.

format/mosaic/mosaic.go
```diff
@@ -0,0 +1,170 @@
...
+			d.FieldStruct(childElm.GetName(), func(d *decode.D) {
+				decodeElement(d, childElm)
+			})
+		} else {
+			arrayName := childElm.GetName()
+			d.FieldArray(arrayName, func(d *decode.D) {
+				// loop over siblings as long as they are the same element
+				for d.Pos() < tagEndBit && !d.End() {
+					elm.PeekNextElement(d, &childElm)
+					if childElm.GetName() != arrayName {
```
wader, line 45: I first thought why not compare id:s but that have to have a 'arrayID` var or call `childElm` methods. So maybe keep it this way.

When does this happen btw? master with two different ids that are not singletons? and i guess one can assume that all of the same id are continuously laid out?

format/mosaic/mosaic.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	case *ebml.UTF8:
+		d.FieldUTF8NullFixedLen("value", int(tagSize))
+	case *ebml.Date: // that's not expected in MOSAIC
+		d.FieldRawLen("value", int64(tagSize)*8)
+	case *ebml.Binary:
+		d.FieldRawLen("value", int64(tagSize)*8)
+
+	case *ebml.Master:
+		if tagID == ebml_mosaic.IdxUnrolledID {
```
wader, line 153: Just curious: How come this is special, thinking why not expressed as non-singleton "KeyOffset" master?

wader (commented):
Overall looks good i think 👍 had some comments and thoughts

format/matroska/ebml_matroska/ebml_matroska_gen.go
martinkirch: yes that was a manual run. Re-generating it adds the `singleton` attribute to all these elements, but that won't affect the program and help later

format/mosaic/mosaic.go
martinkirch: indeed matroska has a few elements having `maxOccurs="1"` and `recurring="1"`, so those could not be rendered that way - maybe they should not be flagged "singleton" by the current code generator?

also that works nicely (espacially for arrays) as long as elements are coherent, maybe that happens less in matroska?

format/mosaic/mosaic.go
martinkirch: that happens for example to top elements, like Tile, Index in `fq '.mosaic' format/mosaic/testdata/basic.mosaic`.

> i guess one can assume that all of the same id are continuously laid out?

not in the general case, as IdxUnrolled below

format/mosaic/mosaic.go
martinkirch: sub-elements of a typical IdxUnrolled can be, in that order:
 * Key
 * Offset
 * Key
 * Offset
 * Void
 * Void (when we erase an entry a-posteriori)
 * Key
 * Offset
 * Key
 * GoToConflicts
 * Key
 * Offset
 * ...

that does not go well with our attempt to group siblings in `decodeMaster`.

I could maybe have decoded them in a Pair'd element, but that would not rigorously reflect the binary format.

format/mosaic/mosaic.go
wader: Ah didn't notice the recurring attribute in the schema, that is good to know. Not sure how often it happens in matroska, will do some research later on. But yes sound reasonable it should not be a singleton in those cases, maybe could replace singleton with maxOccurs and recurring and let format decoder code decide what is a singleton etc?

format/mosaic/mosaic.go
wader: I see, yeap agree keep as close to reflecting the binary format as possible

wader (approved):
This looks mergable to me! had one comment about singleton but that can be done later if it make sense. Anything else you want to add or change?

wader:
If you have asciidoctor installed you could do a `make doc/fq.1.adoc doc/fq.1 doc/fq.1.html` and verify/commit `fq.1` and `fq.1.adoc`. No need to add `fq.1.html`, it will get generated and deployed in a CI job

format/mosaic/mosaic.go
martinkirch: that's a good idea, the code generator should just copy such XML attributes. Changing this.

martinkirch:
I've completed and fix and few EBML elements by the way. It required a minor modification of Matroska's tests. I think we're good !

format/ebml/gen/main.go
```diff
@@ -227,6 +266,36 @@ func main() {
...
+			}
+			if c.Recurring == "1" {
+				fmt.Printf("    Recurring: true,\n")
+			}
+			if c.MinVer != "" {
+				fmt.Printf("    MinVer: %s,\n", c.MinVer)
+			}
+			if c.MaxVer != "" {
+				fmt.Printf("    MaxVer: %s,\n", c.MaxVer)
+			}
```
wader, line 298: Ugg the code for this tool is getting quite ugly 😬 but not sure it's worth tidying up, maybe later :)

wader:
Very nice, thanks and hope fq will be useful! what are the main use cases? used while developing or debugging corruption etc?

martinkirch:
Debugging corruption while developing, at least.

We also care to be readable outside of the archive's infrastructure, so a CLI tool with a query language is an interesting alternative to our python/rust API. Although that supposes fq scales to our typical files' size: in the order of 100GB. I'll test that next week.

wader:
👍 

About file size: fq's resource usage should mostly scale by number of decode values and how much memory their actual/symbolic and descriptions uses. The bits "backing" a decode value is only keept track of as a range and are read on-demand if needed. Also maybe good to know that "raw" fields (ex `d.FieldRawLen`) will have an actual value that is on-demand read, usually used for large blobs etc. I've used fq with 60gb mp4 files without much issues, the memory usage and decode time depend mostly on number of packets/samples.

But note that i haven't spend that much time trying to optimize memory usage so ideas are welcome!

martinkirch:
well, it OOM'd on a 70GB file as expected :)

testing on a smaller one (3,6GB [available here](https://softwareheritage.s3.amazonaws.com/content_shards/train-00000-of-00512.mosaic)) had me notice that even a query one the file's header (like ` '.mosaic.container_meta_data.compression_method'` ) does parse the whole file (that took ~5GB of RAM). do you have any hint on how fq could early exit in such case ?

wader:
> well, it OOM'd on a 70GB file as expected :)

😬 any idea roughly how many decode values that would produce?

> testing on a smaller one (3,6GB [available here](https://softwareheritage.s3.amazonaws.com/content_shards/train-00000-of-00512.mosaic)) had me notice that even a query one the file's header (like ` '.mosaic.container_meta_data.compression_method'` ) does parse the whole file (that took ~5GB of RAM). do you have any hint on how fq could early exit in such case ?

Sadly no. At the moment fq does not do any lazy/deferred decoding. I've tried and have some idea how it could be done but all attempts yet have turned out too complicated or made the decode "DSL" not so nice work with. To keep something similar to the current DSL i think the decoding would have to run in a separate co-routine and send back values somehow 🤔 i'm open to explore how it would be done if you have ideas.

There are some workaround that i used if decoding is too slow or uses lots of memory:
- Use the REPL `-i` to do many quries without re-decoding
- Split the format into subformats if it make sense. Ex: mp3 into header, frames then use header decoder if your not interested in frames etc.
- Format option to skip some heavy decoding. Ex: mp4 decoder has a `decode_samples` option if your not interested in sample details or you can manually decode a sample `fq -d mp4 -o decode_samples=false '.tracks[0].samples[123] | aac_frame' fila.m4a`

martinkirch:
> > well, it OOM'd on a 70GB file as expected :)
> 
> 😬 any idea roughly how many decode values that would produce?

That one contains >13 million objects and for each object we have a dozen EBML elements.

Indeed pushing query information in decoders would be very complicated. On second thought, it also occurred to me that EBML has an additional difficulty: the query should start by the root element ("mosaic" in the  `.mosaic.container_meta_data.compression_method`  query) , the one that contains the entire file except the short EBML header.

I'll try to scratch my head about it but... yeah, it's complicated.

wader:
Did some tries with various degrees of long hanging fruits to use less memory but non of these made any significant difference:
- Embed the scalar value into the decode value to get less allocations and possible less fragmentation
- Make symbol and description pure functions based on actual value. A function/closure ref uses quite a lot also.
- Make the range type an interface and have different compact implementations. Turns out an interface value are two 64bit values (same as a range has to be in worst case), doh :)

What could save a lot in some formats i think would be to have some kind of array decode value. But then format decoders would probably have to explicitly use it somehow.

PR #1376, opened 2026-08-25
bmp: Add decoder

### LINKED ISSUE #745: [Feature request] Support image/bmp
https://en.wikipedia.org/wiki/BMP_file_format

Using fq 0.7.0 (linux amd64) (Arch Linux, installed [extra/fq 0.7.0-1](https://archlinux.org/packages/extra/x86_64/fq/) by pacman), I think there's no support for BMP/DIB format currently.

ChrisJr404:
Adds a decoder for BMP images, closes #745.

It decodes the file header and the BITMAPINFOHEADER family of DIB headers (the common 40 byte one plus the v4/v5 extensions, whose extra bytes go into a `rest` field), the color table for palettized images, and the rest as raw pixel data. Endian is little as usual for BMP.

Tests cover a 24 bit image, a 1 bit palettized image and a BITMAPV4HEADER image. I left the old 12 byte BITMAPCOREHEADER out since I've never run into one, happy to add it if you want.


format/bmp/bmp.go
```diff
@@ -0,0 +1,110 @@
...
+					d.FieldU8("blue")
+					d.FieldU8("green")
+					d.FieldU8("red")
+					d.FieldU8("reserved")
+				})
+			}
+		})
+	}
+
+	if gap := int64(bitmapOffset)*8 - d.Pos(); gap > 0 {
```
wader, line 103: You can try remove this, gaps will be added a fields automatically

format/bmp/bmp.go
```diff
@@ -0,0 +1,110 @@
...
+
+func bmpDecode(d *decode.D) any {
+	d.Endian = decode.LittleEndian
+
+	var bitmapOffset uint64
+	var bitsPerPixel uint64
+	var colorsUsed uint64
+
+	d.FieldStruct("file_header", func(d *decode.D) {
+		d.FieldUTF8("type", 2, d.StrAssert("BM"))
```
wader, line 61: Being in the probe group require that it can be detected without high risk of false positives and not doing much reading. Maybe check for "BM" and make sure file length is at least size big? anything else?

format/bmp/testdata/v4.fqtest
```diff
@@ -0,0 +1,29 @@
+# BITMAPV4HEADER (108 byte) 2x2 32-bit BGRA image, generated by hand
+$ fq -d bmp dv v4.bmp
```
wader, line 2: maybe have one test that don't use -d to test probe

format/bmp/bmp.go
```diff
@@ -0,0 +1,110 @@
...
+	compressionCMYKRLE8       = 12
+	compressionCMYKRLE4       = 13
+)
+
+var compressionNames = scalar.UintMapSymStr{
+	compressionRGB:            "rgb",
+	compressionRLE8:           "rle8",
+	compressionRLE4:           "rle4",
+	compressionBitfields:      "bitfields",
+	compressionJPEG:           "jpeg",
```
wader, line 45: jpeg and png could be decoded as subformats but maybe should be done in a separate PR

format/bmp/bmp.go
```diff
@@ -0,0 +1,110 @@
...
+				})
+			}
+		})
+	}
+
+	if gap := int64(bitmapOffset)*8 - d.Pos(); gap > 0 {
+		d.FieldRawLen("gap", gap)
+	}
+
+	d.FieldRawLen("pixels", d.BitsLeft())
```
wader, line 107: Could we know the size of pixels data instead of assuming all remaining? derive from image_size but specs seems to be a bit unclear if there are special values like 0 that has to be handled

wader (commented):
Good start. Added some comments and thoughts

PR #1403, opened 2026-09-19
docs: point the decoder example at the bson format directory

1816586742-stack:
## What

Fix the example path in the "Steps to add new decoder" list.

## Why

The guide said:

```
- Create a directory  `format/<name>`
- Copy some similar decoder, `format/format/bson.go` is quite small, to `format/<name>/<name>.go`
- Cleanup and fill in the register struct, rename `format.BSON` and add it
to `format/format.go` and don't forget to change the string constant.
- Add an import to `format/all/all.go`
```

`format/format/bson.go` does not exist — `format/format` is not a directory. The BSON decoder lives in its own format directory:

```
$ ls format/bson/
bson.go  bson.jq  bson.md
```

That also matches what the same line says a few words later: follow the `format/<name>/<name>.go` scheme, in the directory created by the step above it. `format/format.go`, referenced on the next line, does exist, so only the example path was wrong.

## Verification

- `format/bson/bson.go` exists; `format/format/` does not
- `format/format.go` exists, so the reference on the following line is unchanged
- scanned every path reference in the repository's markdown: none are stale
- one line changed, no wording edits


wader:
Thanks!

